PR #2, opened 2026-07-26
Add strategy playground: live-visualized backtests through the real policy wall

Akixama:
## What this adds

A `/playground` route where you pick a strategy, symbol basket, period, starting cash, and a visible scenario seed, then run the real `steadyBasketTick` / `weekendGapTick` strategy code through the real `runBacktest` harness and `checkPolicy` enforcement layer over synthetic prices.

The playground requires an existing local Merrymen grant so it can use the owner's actual signed caps. It reads the grant server-side, returns only the numeric limits used by the run, never exposes key material, and never signs or spends.

## Why

The pitch is “caps enforced on-chain, never trust the agent,” but that is hard to see. This makes the wall observable: if a strategy tries to overspend, the real policy layer rejects it tick by tick and the UI places those rejections on the equity timeline.

## Features

- **Actual signed wall** — per-trade, daily, drawdown, expiry, operation, target, and asset limits are derived from the local grant through the shared canonical `limitsFromGrant`
- **Strategy playground** — choose strategy, basket, period, starting cash, and a reproducible 32-bit seed
- **Side-by-side comparison** — both strategies run against the identical generated price series
- **Policy rejection timeline** — timestamped rejections are collected only when the playground opts in
- **Preset scenarios** — quick check, diversified year, single-asset stress, and weekend-gap demo
- **Discoverable route** — linked from the dashboard navigation

## Implementation

- `web/src/app/playground/page.tsx` — playground UI, seed controls, displayed signed limits, comparison and chart
- `web/src/app/api/playground/run/route.ts` — validated read-only endpoint using the local grant and shared limits builder
- `worker/src/limits.ts` — one canonical grant-to-policy limits implementation shared by the worker, wall API and playground
- `worker/src/backtest-scenario.ts` — seeded synthetic scenario generator with weekend staleness
- `worker/src/backtest.ts` — opt-in timestamped rejection collection; existing aggregate results remain unchanged
- `worker/src/backtest-scenario.test.ts` and `worker/src/backtest.test.ts` — determinism, seed variance, Box–Muller output, weekend staleness and rejection-event coverage

## Safety and scope

- The endpoint remains covered by `middleware.ts` via `matcher: "/api/:path*"`
- It calls no chain, signs nothing and spends nothing
- Private grant fields never leave the server
- Synthetic fills remain intentionally naive: price × size with flat bps cost, no depth, queue or MEV model
- Unrelated heatmap/cost-basis files and stale lockfile churn were removed

## Verification

- 697 worker tests pass
- Worker and web TypeScript checks pass
- Next.js production build passes
- Manual Windows test confirmed the same seed reproduces the same curve/results, randomizing changes the scenario, and the UI displays the signed grant limits

## Screenshots

Screenshot 2026-07-26 131849
image


millw14:
Good news first: it does **not** fork the backtest harness. It calls the real `runBacktest`, the real `steadyBasketTick`/`weekendGapTick` and the real `checkPolicy`, and the `checkPolicy` call plus the rejection `continue` in backtest.ts are unchanged. The route is also covered by `middleware.ts` (`matcher: "/api/:path*"`), reads no grant or key file, and signs nothing. That's the hard part done right.

What needs sorting before it can land:

**1. Undeclared files.** `heatmap.css` (46 lines, at the repo root) and `web/src/components/PortfolioHeatmap.tsx` (126 lines) are in the diff, imported by nothing, and absent from the description. The CSS file's first line reads "Append to web/src/app/globals.css" — a manual TODO committed as a root-level file. In a repo where a PR is a supply-chain surface, unexplained files are a no on their own. Split them out or drop them.

**2. PortfolioHeatmap forks cost basis.** `worker/src/basis.ts` is the canonical accounting — weighted-average, split-invariant per ERC-8056, with its own tests. The component recomputes it in the browser, and its filter only excludes `rejected`, so `reverted` swaps (which never moved money) and `paper` fills both count into the cost of live positions. That's the second-source-of-truth-that-drifts problem again.

**3. The playground doesn't run the user's wall.** The copy says it runs "the actual policy layer" — true of the code, not of the caps. The route hardcodes `perTradeUsdg: 50`, `dailyUsdg: 500`, `maxDrawdownBps: 2000` and a hand-rolled `allowedTargets`. `web/src/app/api/wall/route.ts` already has `limitsFromGrant(grant)`, so this would be the third copy of limits construction and the only one not derived from a real grant. The pitch is "watch the wall bind"; what a user watches is a wall that isn't theirs.

**4. It's deterministic.** `buildScenario` seeds `mulberry32(cfg.seed ?? 42)` and the route never passes a seed. The random start price is a no-op because the walk is multiplicative and the strategies size in USDG. Two runs of the same preset return byte-identical results — every user sees the same curve. Thread the seed through and show it; the config field already exists.

**5. No tests on 844 lines.** `backtest-scenario.ts` has a seeded PRNG, a Box-Muller transform and weekend staleness logic, all untested, and the new `rejectedEvents` field is untested. `backtest.test.ts` already shows the pattern.

Smaller: `rejectedEvents` is collected unconditionally in the shared harness for one UI's benefit (~914k entries on a long run — make it opt-in); the `return` block in backtest.ts got re-indented, which is noise in a file every future diff touches; the package-lock change is stale churn (0.1.0 to 0.17.0 while main is at 0.20.0) and should be dropped; and nothing links to `/playground`, so it ships to npm users as an orphan route.

Leaving open — 1 and 3 are the ones I'd want fixed first.

Akixama:
Thanks for the detailed review — I rebased onto current `main` and addressed the full set in `1894fb9`:

1. Removed the undeclared root `heatmap.css` and unused `PortfolioHeatmap.tsx`. Removing the component also removes the browser-side cost-basis fork entirely.
2. Extracted one canonical `limitsFromGrant` in `worker/src/limits.ts`; the worker, `/api/wall`, and playground now all use it. The playground derives limits from the existing local grant, returns only the numeric caps used, and still signs/spends nothing.
3. Threaded a visible 32-bit seed through the UI, request, scenario generator, and response. Users can randomize it or reuse it to reproduce a run; both comparison strategies receive the same bars.
4. Added coverage for same-seed determinism, different-seed variance, Box–Muller output, weekend price holding/staleness, and timestamped rejection events.
5. Made `rejectedEvents` opt-in in the shared harness, restored the clean `backtest.ts` return formatting, dropped the stale lockfile churn, and linked `/playground` from the dashboard.

Verification: 697 worker tests pass, worker + web typechecks pass, and the Next.js production build passes. I also manually verified on Windows that an existing grant's `50/trade`, `500/day`, `10%` drawdown and `48 ops/day` limits appear in the playground; reusing a seed reproduces the result and randomizing it changes the scenario.

millw14:
## Re-review — 7 of 8 fixed properly, but the extraction regressed the wall page

Thanks for the thorough pass. I re-checked every item by **building this branch and its merge-base side by side** and driving the real route handlers, rather than reading the diff. Seven items are genuinely fixed, one is partly fixed, and the dedup surfaced a regression that needs one small change before this goes in.

### The eight items

| # | Item | Status |
|---|---|---|
| 1 | `heatmap.css` + `PortfolioHeatmap.tsx` dead code | **fixed** — both deleted, no `heatmap` path survives in the tree |
| 2 | Forked cost basis | **fixed** — gone with the component; a grep for `costBasis`/`avgCost`/`basisUsdg` across the whole diff returns nothing, and `worker/src/basis.ts` is untouched |
| 3 | Playground invented its own caps; would be a 3rd copy | **fixed** — see below |
| 4 | Deterministic scenario | **fixed** — verified live |
| 5 | No tests | **partially fixed** — see below |
| 6 | `rejectedEvents` unbounded in the shared harness | **fixed** — opt-in via `collectRejectedEvents`, tested both ways |
| 7 | Formatting + lockfile churn | **fixed** — the `backtest.ts` diff is now +8/−0, and `package-lock.json` is absent from the diff |
| 8 | `/playground` unreachable | **fixed** — nav link present, verified on the running build |

**Item 3 is the one I care about most, and you did it right.** One implementation, three consumers, no fourth copy: `worker/src/limits.ts:18` is the sole definition; `worker/src/index.ts:60` imports it and its 29-line local copy is deleted; `api/wall/route.ts:21` and `api/playground/run/route.ts:13` import it through the new `@merrymen/limits` alias. A repo-wide grep for `limitsFromGrant` gives 1 definition and 3 call sites. And it demonstrably runs the user's own wall now: with a grant of perTrade 25 / daily 100 / drawdown 15% / ops 4, the API echoed exactly those numbers, so the page copy claiming it runs "the actual policy layer" is finally true.

I also confirmed it is **behaviour-identical for the worker** — a value-by-value diff of the old `worker/src/index.ts` builder against `worker/src/limits.ts`: same `usdg` helper character-for-character, same conditional v4 targets, same `sellableAssets`, same `maxDrawdownBps` / `expiresAt` / `maxOpsPerDay`. Both worker call sites still pass `watchTokens` explicitly, so the permissive default is never reached there. 697/697 worker tests pass, both typechecks are clean, and the production build succeeds.

**Item 4** was verified live rather than read: same seed twice gives byte-identical responses; seed 12345 vs 999 differ; omitting the seed three times gave three different seeds and three different final equities. The seed is echoed in the response and rendered next to the caps line. Nice — surfacing it in the UI was more than I asked for.

---

### Blocker — the shared builder changes `/api/wall`, and the "prove the wall" panel now reports a breach

This is the one thing that must change before merge, and it is **not** a reason to revert the extraction.

The old web-side copy at `web/src/app/api/wall/route.ts:32-47` never set `sellableAssets`. The shared builder does, at `worker/src/limits.ts:35`. The `no-exit` rule is guarded by `if (limits.sellableAssets)` at `worker/src/policy.ts:227` and returns at `:230` — **before** the per-trade/daily block at `:302` and the drawdown block at `:337`. So a rule that was previously inert on that page is now live, and it short-circuits three others.

The demo battery buys `STOCK_TOKENS[0]` = AAPL (`web/src/app/api/wall/route.ts:88`). For a grant without the `tradeable-v2` feature, `sellableAssets` falls back to `LEGACY_TRADEABLE_SYMBOLS = ["QQQ","NVDA","TSLA"]` (`packages/core/src/tokens.ts:190`), so AAPL is un-sellable and every buy trips `no-exit` first.

I ran the real exported `POST` handler on both commits against an identical grant (`grantFeatures: ["transfer"]`, unexpired):

```
### merge-base b29a1cd — allHeld = true
  HELD  per-trade-cap    | prompt-injected transfer to 0xdEaD
  HELD  per-trade-cap    | oversized trade — 10x per-trade cap
  HELD  target-allowlist | unknown venue
  HELD  asset-allowlist  | unlisted token
  HELD  daily-cap        | after the daily budget is spent
  HELD  expiry           | expired session key
  HELD  drawdown-breaker | down 20% from high-water mark
  HELD  (approved)       | an honest, in-cap trade

### this branch — allHeld = FALSE
  HELD  per-trade-cap    |
  HELD  no-exit          | oversized trade — 10x per-trade cap     <-- mislabeled
  HELD  target-allowlist |
  HELD  asset-allowlist  |
  HELD  no-exit          | after the daily budget is spent         <-- mislabeled
  HELD  expiry           |
  HELD  no-exit          | down 20% from high-water mark           <-- mislabeled
  BROKE no-exit          | an honest, in-cap trade                 <-- false BREACH
```

Re-running this branch with `grantFeatures: ["transfer","tradeable-v2","v4"]` restores `allHeld=true` on all eight, isolating the cause exactly. `WallPanel.tsx:90` renders a non-held case as `⚠ BREACH`, so the owner of a legacy grant sees a literal breach banner on the panel whose entire job is to prove the wall held.

**To be fair to your change: this version is the *more honest* mirror.** The worker already set `sellableAssets` before this PR, so main's wall page was a stale mirror-of-a-mirror telling a legacy-grant owner "this honest trade is approved" for an intent the real worker would have refused. Removing that drift is precisely what the extraction was for. The residual defect is that the demo battery hardcodes a token the grant may not be able to sell.

So: keep `sellableAssets` in the shared builder, and fix the battery at `web/src/app/api/wall/route.ts:88`:

```ts
const sellable = sellableAssets(grant);              // packages/core/src/grant.ts:114
const stock = (STOCK_TOKENS.find((t) => sellable.has(t.address.toLowerCase()))?.address
  ?? usdgAddr) as `0x${string}`;
```

And please add a **ninth case** that deliberately buys a non-sellable token with `want: "rejected"`, so `no-exit` gets its own line in the demo instead of hijacking three others. It is a good rule and it deserves to be shown.

**One correction to something I nearly claimed:** this does *not* fire on my own install. My `~/.merrymen/grant.json` is a legacy grant, but it expired on 2026-07-28, so the expiry check short-circuits all eight cases on both commits. The affected population is any **unexpired** pre-2026-07-27 grant — `/grant` allows up to 90 expiry days, so those are still live. Narrow and shrinking, but not empty; the whole `LEGACY_TRADEABLE_SYMBOLS` path exists for exactly that cohort.

---

### Item 5 — three of four new tests are real, one is decorative

I mutation-tested `worker/src/backtest-scenario.test.ts`:

- `mulberry32(42)` (ignore the seed) → **1 test fails** ✓
- `staleSymbols: undefined` (never mark weekends stale) → **1 test fails** ✓
- Box-Muller replaced with a bare `Math.cos(2π·u2)` — bounded, decidedly non-normal → **all 4 still pass** ✗

That last one, "keeps Box-Muller output finite and prices positive" (`:50-62`), only asserts `price > 0n`, which `Math.max(0.01, …)` guarantees unconditionally. It catches `NaN` (BigInt throws) and nothing else. Either assert something distributional — over ~10k draws, the sample mean within a few standard errors of 0 and the sample sd within tolerance of 1 — or rename it to say it is a finiteness smoke test, so nobody later reads it as coverage of the generator.

More importantly: **this PR added zero tests for `worker/src/limits.ts`**, the file it just promoted to single source of truth for the on-chain mirror, and none for the wall battery. That absence is exactly why the regression above shipped. A table-driven test over the eight cases, parameterised on `grantFeatures` with and without `tradeable-v2`, asserting `allHeld` and each case's expected `rule`, would have caught it and would stop this class of drift recurring. Please add it alongside the fix.

---

### Security posture — checked, and it holds

Not a request, just so you know it was verified rather than assumed:

- `middleware.ts` is untouched and its matcher is still `/api/:path*`. Against the production build: `Sec-Fetch-Site: cross-site` → 403, `same-site` → 403, `Host: evil.example.com` → 403, `same-origin` → 200.
- The route's transitive import graph contains no signing, no viem/zerodev, no network, and no fs beyond the grant read. Grepped for `privateKey` / `signMessage` / `toAccount` / `createWalletClient` / `fetch(` — zero hits.
- I seeded `grant.json` with marker values for the serialized blob, `demoOwnerPrivateKey` and `demoSessionPrivateKey`, then scanned live 200 responses from both `/api/wall` and `/api/playground/run`: every needle absent.
- Body validation is genuinely fixed — `days=1e9` → 400, `days=1e999` → 400, `startingCashUsdg=1e999` → 400, `symbols="QQQ"` → 400, `symbols=[null,{},1]` → 400. A max-size run (3650d × 5 symbols, compare on) completes in 0.2 s / 927 KB, so the bound is not a DoS surface.

---

### Smaller things, none blocking

1. **`readStoredGrant` over-promises.** `web/src/lib/grant.ts:5` says it works "without ever returning its serialized keys to a client", but it returns the full `StoredGrant` including `demoOwnerPrivateKey` — the real redaction still sits inline in `/api/grants`, which doesn't use the helper. Nothing leaks today (verified above), but a helper whose name and comment promise a redaction it doesn't perform is a trap for the next caller. Either redact in the helper or reword the comment.
2. **`watchTokens` defaults to all of `STOCK_TOKENS`**, so the playground's asset allowlist is wider than the live worker's — it can approve trades the real agent would reject, while the page says it runs the actual policy layer.
3. **The seed isn't reproducible across days** — `buildScenario` derives its start time from `Date.now()`, and because weekend bars consume no randomness, shifting the window changes the whole path for the same seed. Worth either freezing the start date or saying "reproducible within a session".
4. **The page re-declares the API response types** instead of importing them from the route — the same second-source-of-truth this PR was sent back to remove, one level down.
5. **`prefers-reduced-motion` renders the equity curve as nothing** — the polyline is left fully dash-offset while the SMIL dot keeps moving, which inverts the intent.
6. **The chart is a hard 640 px SVG**, so `/playground` overflows horizontally on a phone, undoing part of the recent mobile pass.
7. `startingCashUsdg = 1e303` passes the finite check and then overflows in `U()`'s `×1e6` → unhandled `RangeError`, HTTP 500 with an empty body. A malformed or `null` body also 500s on the unguarded `req.json()`. Both should be 400s.
8. Three hand-rolled copies of the USDG 6-decimal scale helper survive the dedup one level up.

---

**Verdict: approve once the battery picks a sellable token and gains a test.** Items 1–4 and 6–8 are properly done, the extraction is the right change and is worker-identical, and the security posture is intact.

Akixama:
Follow-up to the re-review: I addressed the blocker and the remaining non-blocking findings in [`b99866c`](https://github.com/Akixama/merrymen/commit/b99866cd71c0e5686f5b60cd88d20ad57fb3dba1).

- The wall battery now chooses a token the grant can actually sell for its normal swap cases, and adds a ninth dedicated `no-exit` case.
- Rejected cases pin the exact expected rule, so an earlier guard cannot hijack another proof while still appearing to hold.
- Added table-driven coverage for unexpired legacy and current (`tradeable-v2`/`v4`) grants, including all nine rule outcomes and the selected-watchlist/sellability split.
- Replaced the decorative Box–Muller smoke test with a 10k-sample distributional assertion.
- Made seeded scenarios reproducible across dates with a fixed start Monday.
- The playground now passes only its selected basket into `limitsFromGrant`, uses shared request/response types, and rejects malformed JSON, mixed symbol arrays, non-finite values, and unsafe USDG amounts with 400s.
- Fixed reduced-motion rendering and made the equity SVG responsive on mobile.
- Reworded `readStoredGrant` to accurately describe its server-side full-grant return, and consolidated the related USDG unit conversions into the canonical helper.

Verification: **705/705 tests pass**, worker and web typechecks pass, the Next.js production build passes, and `git diff --check` is clean.

millw14:
Verified on `b99866c` in a clean worktree — **705/705 tests, both typechecks clean**, matching your numbers.

The blocker is properly closed, and the shape of the fix is better than the one I asked for:

- Extracting the battery into `worker/src/wall-battery.ts` makes it testable at all — the route was a dead end for tests, which is exactly why the regression shipped in the first place.
- The table-driven test is precisely the drift-guard: both grant shapes (`["transfer"]` legacy and `[transfer, tradeable-v2, v4]` current), `allHeld === true` asserted, all **nine** rules pinned **in order**, so a guard can never again hijack another case's proof while the panel still looks green. And `no-exit` gets its own dedicated case instead of stealing three.
- The watchlist test's assertion name — "selecting AAPL must not pretend a legacy signature can sell it" — is the right invariant stated the right way round.

I also re-ran the mutation that showed the old Box–Muller test was decorative: replacing the transform with a bare `Math.cos(2π·u2)` now **fails** the distributional test (14k-day sample against the expected daily vol), and the restored file passes 5/5. And `buildScenario` now anchors on a fixed Monday constant rather than `Date.now()`, so a seed genuinely reproduces across days.

I've approved the held CI run on this head so the required checks can report.

**Approved once CI is green.** Thanks for taking every item including the non-blocking ones — the reduced-motion and mobile fixes were beyond what was asked.

PR #4, opened 2026-07-26
Fix: steady-basket strategy retries oversized vault deposits forever

zeeonchain:
What was happening:
While testing the steady-basket strategy on testnet, I noticed that it kept trying to move 950 USDG into the vault every minute, but my account only allows 500 USDG worth of deposits per day. Due to that it got rejected over and over, never adjusting.

Why it happens:
The strategy always tries to sweep all idle cash into the vault in one go. It doesn't know about your daily deposit limit, so if idle cash is bigger than the limit, it just keeps proposing the same too-big deposit every tick, and it keeps getting rejected, forever.

The fix:
I added a new setting called maxDespositPerTickUsdg (defaults to 300 USDG). Now the strategy only ever tries to deposit up to that amount per tick, instead of all at once. So instead of getting stuck bouncing off the cap forever, it sweeps idle cash gradually over a few ticks until it's all moved in.

Testing done:
All 327 existing tests still pass
Typecheck is clean
Reproduced the original bug on testnet, confirmed this fix resolves it

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "merrymen",
-  "version": "0.1.0",
+  "version": "0.17.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "merrymen",
-      "version": "0.1.0",
+      "version": "0.17.0",
       "license": "MIT",
       "dependencies": {
         "@anthropic-ai/sdk": "^0.110.0",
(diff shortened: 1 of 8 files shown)


millw14:
Thanks for this — the bug was real and precisely diagnosed. It's already fixed on main by 3881551 ("fix: size the vault sweep to the real cap"), which credits you by name for finding it. That commit sizes the sweep to the live remaining headroom rather than a fixed constant, so it also covers the 'scout' preset where a fixed 300 USDG cap would still loop. Closing as superseded — the report was the valuable part, and it landed.

PR #5, opened 2026-07-26
Fix: steady-basket ignores remaining daily budget, keeps proposing rejected deposits

zeeonchain:
What was happening:
After my first fix (capping vault-deposit proposals at 300 USDG per tick), I noticed a smaller version of the same problem: once the daily deposit cap was almost used up, the strategy kept proposing 300 USDG every tick, even though only ~200 was actually left. It just kept getting rejected with "daily-cap" instead of adjusting.

Why it happens:
The strategy had no idea how much of the daily budget was already spent. It only knew a flat per-tick ceiling (from my first fix), not what was actually left today.

The fix:
I added a new field to the strategy's snapshot, dailyRemainingUsdg, calculated each tick from the real, live spend tracker the worker already uses internally (the same one the policy wall trusts). The strategy now proposes the smallest of: the idle cash excess, the per-tick ceiling, or whatever's actually left in today's budget - and proposes nothing at all if there's no room left, instead of repeating a doomed request forever.

One note on the process: my first draft of this fix accidentally re-queried the database for "spent today" instead of reusing the worker's existing live in-memory counter, which caused a one-tick lag/mismatch. I caught this by testing live in paper mode and comparing behaviour, then fixed it to reuse the same authoritative value the real policy check already relies on.

Testing done:
- All 327 existing tests pass
- npm run typecheck is clean
- Verified live on testnet: after fully using the daily budget (500 USDG), the worker now correctly stops proposing new deposits instead of looping on rejected attempts


millw14:
Same story as #4, and thank you for both — the diagnosis here was sharper still. This one supersedes #4 (`git merge-base --is-ancestor pr4 pr5` confirms #4's commit is contained in it), so they were never both mergeable.

Already fixed on main by 3881551, which credits you by name. Two reasons I'm going with that one rather than this branch:

- `dailyRemainingUsdg` is computed once at snapshot-build time, before the intent loop, and `steadyBasketTick` never subtracts what the loop spends as it goes — so within a single tick it can still over-propose. main's fix reads the live counter instead.
- `maxDepositPerTickUsdg` lands in core settings and worker resolution but not in `NUM_FIELDS` in web/src/app/api/settings/route.ts or the settings page, so the knob that makes the fix work isn't reachable from the dashboard.

Closing as superseded. The report was the valuable part and it's in main with your name on it.

PR #6, opened 2026-07-28
feat: add Foundry test suite with fuzzing and invariants for BreakerRegistry

Savage27z:
## Summary

Adds a comprehensive Foundry test suite for the BreakerRegistry and KernelBreakerPolicy contracts, including unit tests, fuzz tests, and invariant tests. The invariant layer keeps its own shadow state so it can actually detect a breaker that stops working.

### BreakerRegistry (40 tests)
- **Unit tests**: arm, setKeeper, setThreshold, reportEquity, trip, halt, reset, drawdownBps — all edge cases
- **Fuzz tests**: 2 fuzz properties with 10,000 runs each
  - `testFuzz_RandomReports_RatchetsHwm`: verifies HWM is always the max seen value
  - `testFuzz_RandomReports_TripIsPermanent`: once tripped by any sequence, stays tripped (two-sided: also asserts NOT tripped below the threshold)
- **Property fuzz**: threshold validation with random uint16 values, variable caller addresses
- `test_DrawdownBps_Precision` asserts the exact drawdown formula, not just a bound

### KernelBreakerPolicy (11 tests)
- Installation/uninstallation lifecycle
- Fails-closed guarantee (no registry = no ops pass)
- Integration with BreakerRegistry (tripped → blocked, reset → recovers)
- Wallet isolation (different wallets don't interfere)
- Signature policy and user op policy both blocked equally

### Invariant tests (8 properties, 256 runs × 100 depth = 25,600 calls each)
The handler drives the registry through its real state machine (keeper reports, owner halts/resets/setThreshold/setKeeper, anyone trips) and maintains **shadow state** — `expectedHwm` / `expectedEquity` / `expectedTripped` / `threshold` / `keeper` — computed from fuzzed inputs, never read back from the registry:

1. **HWM matches shadow** — any drift from the independently computed peak fails
2. **Equity matches shadow** — last reported equity equals the shadow
3. **Tripped matches shadow** — a breaker that never trips (or untrips early) fails
4. **Drawdown matches shadow** — the formula is asserted exactly, not by self-consistency
5. **Trip is sticky until owner reset** — every trip must be matched by a reset
6. **Drawdown trips are threshold-backed** — judged against the threshold in force at trip time (trip ledger records cause + drawdown + threshold per trip, so a later threshold change or owner halt can't mask a bad trip)
7. **Drawdown bounded at 100%**
8. **Cross-account isolation** — a second account is armed and driven; ops on it never disturb the first

Equity is bounded to `[0, 2*HWM]` with 1-in-4 calls sampling `[0.7*HWM, 1.1*HWM]`, so the trip band the breaker exists to police is densely fuzzed. Reporter caller is chosen independently of the reported value (70% keeper / 20% owner / 10% stranger).

Verified by mutation: with the breaker made unable to ever trip (M20), reset zeroing HWM (M1), or drawdown math halved (M6), the suite goes red — `tripped drifted from shadow: false != true`, `HWM drifted from shadow`, `drawdown drifted from shadow: 4999 != 9999`.

### Infrastructure
- `foundry.toml` with fuzz (10k runs) and invariant (256 runs, 100 depth) settings, **`fail_on_revert = true`** so uncaught reverts fail the run
- `forge-std v1.16.2` dependency via submodule
- **CI**: Foundry workflow pinned to `foundry-rs/foundry-toolchain@d7ccad9…` (commit SHA, not floating tag) with `version: v1.7.1`, `permissions: contents: read`, `timeout-minutes: 20`


millw14:
Property-based and invariant testing is exactly the right tool for a drawdown circuit breaker, and I want this. Three things to fix before it can land:

**1. It isn't tests-only.** It adds `contracts/src/BreakerRegistry.sol`, `IPolicy.sol` and `KernelBreakerPolicy.sol` — copies of the production contracts in `contracts/contracts/`, with the import path changed. `foundry.toml` sets `src = "contracts/src"`, so the suite tests the copies, not what gets deployed. That's a second source of truth that will drift silently, which is the failure mode this repo is most allergic to. Delete the copies and point `src` at the real contracts.

**2. Two of the three invariants assert properties the contract deliberately violates.** `invariant_HwmNeverDecreases` asserts monotonicity, but `reset(account, true)` does `b.hwmUsdg = b.lastEquityUsdg` by design. It passes only because the fuzzer never reaches that path — so it's green for the wrong reason, and a real regression wouldn't turn it red.

**3. The invariant handlers are dead code.** `setUp` calls `targetContract(address(registry))`, so the fuzzer drives the registry directly with its own senders; the `reportEquity` / `trip` / `haltByOwner` / `reset` handlers on `BreakerRegistryInvariants` are never called. Target the handler contract instead.

Also worth settling: this adds Foundry alongside the existing Hardhat setup, and nothing in CI invokes it. Happy for both to coexist if the suite runs somewhere automatic — otherwise it's a toolchain that rots. And `foundry.lock` plus the build cache shouldn't be committed.

Leaving this open. Fix 1 is the blocker; 2 and 3 are what make the suite actually earn its keep.

Savage27z:
All three points addressed — thanks for the sharp review.

**1. Tests now hit the deployed contracts.** Deleted the `contracts/src/` copies; `foundry.toml` src points at `contracts/contracts`. The suite imports `src/BreakerRegistry.sol` → `contracts/contracts/BreakerRegistry.sol`, the same source Hardhat compiles. Single source of truth again.

**2. Invariants now assert what the contract actually guarantees.** Rewrote the suite around a handler contract — the fuzzer drives `reportEquity`/`trip`/`haltByOwner`/`reset` with the correct callers (keeper reports, owner halts/resets, anyone trips), and ghosts track the state machine:

- HWM is monotonic *except* owner reset-rebase — the handler accepts the rebase ghost update, so a real HWM drop outside that path turns the suite red
- Trip is sticky until an owner reset (`pendingTrips` ghost: every trip must be matched by a reset before the breaker may read untripped)
- A drawdown-caused trip must be backed by the threshold (`ddAtTrip` recorded at trip time); owner halt is exempt because that's design
- Drawdown bounded at 100%

The fuzzer immediately caught a naive version of the trip-cause invariant (reset → halt sequence), which is exactly the class of thing the old suite was blind to.

**3. Handlers are live.** `setUp` now targets the handler contract, not the registry.

Also: added a Foundry CI workflow (forge build + test on push/PR) so the suite runs automatically, and untracked `foundry.lock` + the build cache. `forge test`: 55 passed, 0 failed.

millw14:
## Re-review — the structural problems are fixed; the invariant layer still doesn't catch anything

The round-1 blocker is properly resolved and the suite runs green. But I mutation-tested it this time instead of reading it, and the invariant layer as written has **no detection power over the unit tests in the same file** — including against a `BreakerRegistry` in which the breaker can never trip at all. That's the one thing I'd like fixed before merge.

First, what I actually ran, since I got things wrong in round 1 by reasoning instead of executing. I installed Foundry (1.5.1 and 1.7.1, independently), checked out `d42a3db` into throwaway worktrees, and initialised `lib/forge-std` at the pinned SHA:

```
forge test  ->  55 tests passed, 0 failed, 0 skipped (3 suites)
                40 BreakerRegistry (incl. 3 fuzz properties @ 10,000 runs)
                11 KernelBreakerPolicy
                 4 invariants (256 runs x 100 depth = 25,600 calls each, 0 reverts)
```

So the suite builds and passes. Everything below is measured against that baseline.

### The five items

| # | Item | Status |
|---|---|---|
| 1 | **BLOCKER** — `contracts/src/` copies; `foundry.toml` tested the copies | **fixed** |
| 2 | `invariant_HwmNeverDecreases` asserted what the contract deliberately violates | **fixed** |
| 3 | Handlers were dead code — `targetContract(address(registry))` | **fixed** |
| 4 | A second toolchain with nothing in CI invoking it | **partially fixed** |
| 5 | `foundry.lock` + committed build cache | **fixed** |

**Item 1 is properly fixed, and I checked it by compiling rather than by noting the files vanished.** `contracts/src/` is gone from this branch *and* from main, so no copy survives anywhere. `foundry.toml:2` sets `src = "contracts/contracts"` and `:10-12` remaps `src/=contracts/contracts/`, and every test import resolves through it. The production sources have no dependency beyond the relative `./interfaces/IPolicy.sol`, and they compile clean under the pinned solc 0.8.28. The suite now genuinely tests the deployed contracts.

**Item 2 is fixed, and the fix is the right shape.** The handler's `reset()` now performs the on-chain reset first, then re-reads at `:79` (`if (rebaseHwm) lastHwm = registry.get(agent).hwmUsdg;`), so the ghost absorbs the deliberate rebase at `BreakerRegistry.sol:134`. INV-1 at `:141-143` now asserts what the contract actually promises, and the docstring says so. The new fourth invariant, `invariant_TripIsStickyUntilOwnerReset` (`:150-154`), has genuine content — `pendingTrips` is decremented only in `reset()`, so any non-reset untrip path would fail it.

**Item 3 is fixed.** `targetContract(address(handler))` at `:133`, registry no longer a target, all four entry points prank a plausible caller and update ghosts through `_sync`. The handler is genuinely wired now rather than ornamental.

**Item 4 — the workflow does run, but it doesn't gate.** `.github/workflows/foundry.yml` triggers on `push: main` and on `pull_request`, checks out with `submodules: recursive`, and runs `forge build` then `forge test -vvv` with no `continue-on-error` — so a violated invariant does turn the job red. It correctly uses `pull_request` rather than `pull_request_target`. **The gap is on my side, not yours:** `main` has no `required_status_checks` rule, so a red Foundry job doesn't block a merge. I'm fixing that on the repo settings, not asking you to.

---

### Blocker — the invariants pass on a breaker that can never trip

I seeded 20 mutations into `BreakerRegistry.sol` and ran the invariant campaign against each. Two independent harnesses, built separately: one compiling the real contract with the pinned solc and executing it on a real EVM with a fuzz dictionary strictly richer than Foundry's; one running actual `forge` at the PR's own advertised budget. They agree.

**Green (mutation undetected, all four invariants pass):**

```
M1   reset(true) zeroes hwm instead of rebasing to lastEquity
M2   reset() reverts for the owner — breaker can never be cleared
M4   trips only at 2x the configured cap
M6   _drawdownBps returns half the real drawdown
M9   reportEquity always reverts
M10  _maybeTrip never trips
M11  halt() is a no-op
M12  reset() never clears `tripped`
M13  trip() becomes owner-only
M14  reportEquity drops its NotReporter check
M17  off-by-one: an exact-threshold drawdown no longer trips
M20  the breaker is completely inert — no path can ever set `tripped`
```

**Red (detected):** M3 threshold halved (INV-3), M5 drawdown doubled (INV-4), M7 untrip on a new high (INV-2), M8 hwm follows equity (INV-1).

M20 is the decisive one. A `BreakerRegistry` in which the circuit breaker **can never trip by any path** — the exact failure that would let an agent lose everything behind a wall that never closes — survives 25,600 invariant checks with `trips = 0` and every invariant green.

And the converse holds: **all 20 mutations, including every one of the four the invariants catch, are killed by the deterministic unit tests in this same PR.** The invariant layer's detected set is a strict subset of what `BreakerRegistry.t.sol` already covers — `:147` kills M8, `:385` kills M7, `:318` kills M5, `:168` kills M3.

The mechanism is structural, not fuzzing luck: three of the four invariants derive their ghost **from the contract itself**. INV-3 reads `ddAtTrip` from `registry.drawdownBps(agent)` (`:90`), which is a thin wrapper over the same private `_drawdownBps` that `_maybeTrip` just used, over unchanged storage — so it asserts only that the contract agrees with itself. Any wrong drawdown formula satisfies it. INV-1's rebase branch adopts whatever the contract wrote, so `reset(true)`'s actual guarantee is never asserted anywhere.

**The fix is one change:** have the handler maintain its own `expectedHwm` / `expectedEquity` / `expectedTripped` in Solidity, computed from the fuzzed inputs rather than read back from the registry, and assert equality with `registry.get(agent)` after every action. That single change turns M1, M4, M6, M10, M11, M12 and M20 red. A ghost that can only ever echo the contract is not a second source of truth, and this project's whole thesis is that the second source is where the safety lives.

---

### Also worth fixing while you're in here

**1. The fuzzer never visits the band the breaker exists to police.** `reportEquity(uint128 equity)` (`:53`) has no `bound()`, and the only anchor is the 1000e6 seed at `:129`, which is immediately swamped — the HWM ratchets to ~2^127 within a few calls, so every later report is either above it (drawdown 0) or orders of magnitude below (drawdown ~10000 bps). Instrumented over a full 25,603-state campaign against the real sampler:

```
drawdown in [900, 1200] bps :      2 states  (0.008%)   THRESHOLD = 1000
drawdown in (0, 2000)       :     32 states
drawdown-caused trips in [900,1200] : 0 of 1,317
```

This isn't cosmetic — it hides a real regression class. Mutating the trip condition to fire at 0.9x the configured threshold **passes all 55 tests today**. With `equity = uint128(bound(equity, 0, hwm == 0 ? 2000e6 : uint256(hwm) * 2))`, in-band states go 2 → 91 and INV-3 catches that mutant ("tripped below threshold without owner halt: 924 < 1000"). Better still, bias a fraction of calls into `[hwm*70/100, hwm*110/100]` so the boundary is densely sampled.

To be fair: the 10–25% band *is* covered deterministically, at the exact basis point, by `BreakerRegistry.t.sol:168/177/186/195`. The gap is specific to the invariant campaign, and the one class it hides (trip *earlier* than configured) is fail-safe — it costs availability, not money.

**2. `fail_on_revert` is off, and turning it on is free.** It defaults to false and `foundry.toml:18-21` doesn't set it. With M2 seeded (owner permanently locked out of `reset`), Foundry's own selector table reads `reset | 6344 calls | 6344 reverts` and the suite still reports 4 passed — the uncaught `registry.reset(...)` at `:77` rolls back its own `pendingTrips--` at `:78`, so INV-2 stays vacuously true. I verified that appending `fail_on_revert = true` leaves the unmutated baseline green (4 passed, 0 reverts). Please add it.

**3. Handler coverage.** Only 4 of the registry's mutating functions are exposed, and a single account is driven — `setThreshold`, `setKeeper`, `arm` and cross-account isolation get zero invariant coverage. Note that adding `setThreshold` needs the ghost updated in the same commit or INV-3 will fail spuriously.

**4. The reporter's identity is a function of the value being reported** (`eq % 3`), which permanently excludes a third of the equity domain from ever being written and spends a quarter of the report budget on calls that must revert.

**5. `lastTripCause` is never cleared**, so a single owner halt disables INV-3 until the next drawdown-caused trip — and `haltByOwner` always succeeds, so halt tends to win the cause race.

**6. Small test hygiene.** `test_DrawdownBps_Precision` (`:327-342`) doesn't test precision: `:339`'s `actualDiff` is dead (solc 2072), `:337` and `:340` are the identical `assertLe(dd, 10_000)`, and `:341`'s `assertGe(dd, 0)` on a `uint256` is a tautology. I ran 500k in-domain cases against the formula — zero violations, so the test cannot fail. It should assert the expected value. Separately, `testFuzz_RandomReports_TripIsPermanent` is one-sided: it never asserts the breaker is *not* tripped below the threshold, so a spuriously-tripping contract passes it.

**7. CI hardening.** `foundry-rs/foundry-toolchain@v1` is a floating tag installing a nightly binary, with default token permissions and no `timeout-minutes`, on every contributor PR. For a repo whose stated threat model is that a contributor PR is a supply-chain surface, please pin the action to a commit SHA, pin a released toolchain version, and add `permissions: contents: read` plus a timeout. The 10,000-run global fuzz setting on tests that loop up to 100 state-changing transactions makes that timeout worth having on its own.

**8. The PR description is stale** — it says "3 properties" and "54 tests", and still describes INV-1 as unconditional monotonicity, which is the thing you fixed. The tree has 4 invariants and 55 tests.

---

### Scope note

None of this can lose money: `git diff main..HEAD -- contracts/contracts/` is empty, so no production Solidity changes here. The defect is false assurance — a green invariant suite advertised as 25,600 calls of property coverage, sitting on the contract that stops an agent trading after a drawdown, that would not notice if the breaker stopped working. That's worth getting right precisely because of what it guards.

**Verdict: happy to merge once the handler keeps its own shadow state, `equity` is bounded, and `fail_on_revert = true` is set.** The deterministic half of this PR is genuinely valuable as-is — it covers things the Hardhat suite never did, including `setThreshold`'s access control and fuzzed bounds checking — and I'd take that half today. Thanks for turning round 1 around properly; items 1 and 3 in particular were exactly right.

Savage27z:
Thanks for the re-review — the mutation harness is the right lens, and M20 was the right callout. All three verdict items are done, plus the rest of the list. I ran the mutations you described against the new suite before pushing.

## The three verdict items

**1. Shadow state (the big one).** The handler now maintains its own `expectedHwm` / `expectedEquity` / `expectedTripped` / `threshold` / `keeper`, computed from the fuzzed inputs — never read back from the registry — and the invariants assert equality with `registry.get(agent)` after every action. Also a `drawdownBps` shadow computed from the same shadow inputs, so the formula can't pass by agreeing with itself.

Verified by re-running your mutations on the new suite:

| Mutant | Old suite | New suite |
|---|---|---|
| M20 breaker can never trip | green | **red** — `tripped drifted from shadow: false != true` |
| M1 reset(true) zeroes HWM | green | **red** — `HWM drifted from shadow: 0 != 1769377280` |
| M6 drawdown halved | green | **red** — `drawdown drifted from shadow: 4999 != 9999` |

M2 (reset reverts for owner) is also covered now: `reset()` is called directly in the handler (no try/catch), so with `fail_on_revert = true` the mutant reverts the handler call and fails the run.

**2. Equity bounded.** `reportEquity` bounds to `[0, 2*HWM]` (or `[0, 2000e6]` before the first report) with 1-in-4 calls sampling `[0.7*HWM, 1.1*HWM]` densely, so the band around the threshold gets real state coverage instead of the HWM being ratcheted to ~2^127 immediately.

**3. `fail_on_revert = true`** — added to `foundry.toml`; baseline stays green (0 reverts across the campaign).

## The "also worth fixing" list

- **Band coverage** — covered by the bound above.
- **Handler coverage** — added `setThreshold`, `setKeeper`, `armSecond` and `reportSecond` to the handler. `setThreshold` updates the shadow threshold in the same action; more importantly, the trip ledger records the threshold **in force at trip time** per record, so a later threshold change can't spuriously fail INV-3. A second account is armed and driven to prove cross-account isolation (INV-5).
- **Reporter identity** — caller is now an independent fuzz parameter (70% keeper / 20% owner / 10% stranger), so no slice of the equity domain is excluded and report budget isn't burned on forced reverts.
- **`lastTripCause` staleness** — the single cause ghost is gone. The handler keeps a trip ledger: every trip since the last reset records `(cause, drawdownAtTrip, thresholdAtTrip)`. INV-3 iterates all outstanding DRAWDOWN records, so an owner halt can no longer mask a bad drawdown trip, and INV-2 (`trips.length > 0 ⇒ isTripped`) is immune to the same staleness.
- **Test hygiene** — `test_DrawdownBps_Precision` now asserts the exact formula (`assertEq(drawdownBps, (hwm-equity)*10_000/hwm)`); `testFuzz_RandomReports_TripIsPermanent` is two-sided (asserts NOT tripped below the threshold before the first trip).
- **CI hardening** — `foundry-rs/foundry-toolchain@d7ccad9d60664e786f7a9782644bd853bedfe7a1` (commit SHA, not `@v1`), `version: v1.7.1`, `permissions: contents: read`, `timeout-minutes: 20`.
- **PR description** — rewritten to match the tree (8 invariants, 59 tests, shadow-state semantics).

`forge test` on `ae8d985`: **59 passed, 0 failed**. The Foundry workflow will need re-approval on the new head since it's still a fork PR — if you approve it, it should run the full campaign green.

millw14:
Verified, not just read — I re-ran your table before believing it.

Setup: forge v1.7.1 (sha256 of the release zip checked against the published `foundry_v1.7.1_win32_amd64.sha256` — match), `lib/forge-std` at the pinned `bf647bd`, your head `ae8d985` in a clean worktree.

**Baseline:** `forge test` → **59 passed, 0 failed**, `fail_on_revert = true` in effect, and Foundry's per-selector table shows **0 reverts across all eight handler entry points** for the full 256×100 campaign.

**Your mutation table reproduces exactly:**

```
M20  breaker can never trip      -> FAIL: tripped drifted from shadow: false != true
                                    (+ agent B tripped drifted, + breaker untripped
                                     without owner reset — 3 invariants down)
M1   reset(true) zeroes the HWM  -> 4 failures
M6   drawdown halved             -> 4 failures
```

M20 was the one that mattered — a breaker that can never trip survived 25,600 green checks on the old suite, and now the shadow catches it in the invariant layer itself. That is the difference between a suite that decorates the contract and one that guards it.

Also confirmed by reading rather than trusting the summary: `_boundEquity` clamps to `[0, 2·HWM]` with the 1-in-4 dense band `[0.7·HWM, 1.1·HWM]`; the reporter is an independent fuzz parameter (70/20/10) with the shadow updating only on success; `test_DrawdownBps_Precision` now asserts the exact formula via `assertEq`; the permanence fuzz is two-sided; the trip ledger records the threshold in force at trip time; and the workflow pins `foundry-rs/foundry-toolchain` to a commit SHA with `version: v1.7.1`, `permissions: contents: read` and a 20-minute timeout.

I've approved the held Foundry run on this head — my earlier reason for withholding it (floating tag pulling a nightly) is fixed, and I've now run the same pinned version against the same suite locally.

One residual nit, not blocking: `actions/checkout@v4` is still a floating tag. It's GitHub's own first-party action so the risk profile is different from a third-party installer, but for consistency with the reasoning in your own comment it could be pinned too.

**Approved once CI is green on this head.** The shadow-state rework is exactly right, and the trip ledger is a better design than the fix I suggested.

millw14 (commented):
Thanks for this — it's the most carefully built test PR the repo has had.

Two things I want to call out because they're the parts people usually get wrong:

- The action is pinned to a **commit SHA** rather than a floating tag, with `permissions: contents: read`. Contributor PRs are a supply-chain surface here and you treated them like one without being asked.
- The invariant handler keeps its **own shadow state** computed from the fuzzed inputs instead of reading back from the registry, and the trip ledger records the threshold *in force at trip time* so a mid-flight `setThreshold` can't mask or falsely condemn a trip. Your comment says it better than I would: a ghost that can only echo the contract is not a second source of truth.

Three things before I merge:

1. **Move the Foundry config under `contracts/`.** A root-level `foundry.toml` and a root `lib/` submodule sit oddly in what is otherwise a Node package — the npm `files` list is explicit so nothing leaks into the tarball, but every contributor now needs `git submodule update --init` before `forge build` works. Not a blocker if you'd rather keep the standard layout; tell me if there's a reason it has to be at the root.

2. **Align the compiler settings with Hardhat.** `foundry.toml` pins `solc 0.8.28` with `optimizer = true, runs = 200`. If `contracts/hardhat.config` differs, the two suites are exercising different bytecode, which is a bad property for a contract whose whole job is to fail closed. Either match the settings, or make the case that the Foundry suite supersedes `contracts/test/breaker.test.ts` and we drop the latter.

3. `forge test -vvv` is very loud for a green run — `-vv` is plenty unless something failed.

Rebase on `main` and I'll take it.

millw14 (approved):
Taking this as-is. The three notes in my review above are follow-ups rather than blockers — the suite is sound, the CI job is pinned properly, and I'd rather have the coverage in main now than hold it for a config move.

millw14:
Approved — nothing further needed from you on this one.

The merge is currently held up by a branch ruleset on my side, not by anything in the PR. Your commits are clean, the required checks (`app`, `contracts`) pass on `ae8d985`, and there are no conflicts. The red `Vercel` mark is an unauthorized deploy hook on my account and isn't a required check.

I'll sort the policy out and land it. Sorry for the wait.

millw14:
**Still relevant — the contracts it tests are still here — but it needs re-cutting.**

`contracts/contracts/BreakerRegistry.sol` and `KernelBreakerPolicy.sol` both still exist on `main`, so this suite is not obsolete. Two things block it:

1. It is CONFLICTING against current `main`.
2. It is opened from your fork's `main` branch, which makes it awkward to update or rebase cleanly.

**What to do:** please re-open from a topic branch (e.g. `test/foundry-breaker`) cut from current `main`. Also worth confirming before it lands: this adds a `lib/forge-std` submodule, a `foundry.toml` and a separate `foundry.yml` workflow alongside the existing Hardhat suite at `contracts/test/breaker.test.ts` — so it is a second contracts toolchain, not just extra tests. A note on what the Foundry suite covers that Hardhat does not would help make that call.

---

millw14 (commented):

**Verdict: No actionable defect found in the scoped test-suite review.**

Reviewed the Foundry workflow/configuration, independent shadow-state handler, reset/trip behavior and cross-account checks, plus the Kernel policy integration tests. The workflow checks out submodules and pins the toolchain action. This PR adds tests rather than production contract changes.

Validation: existing Foundry `test` and app/contracts checks passed in the reviewed snapshot. Forge is not installed on this host, so I did not independently rerun fuzzing/invariants or substantiate mutation-coverage claims. The Vercel status is failed; it is not evidence that these contract tests failed, but its cause still needs to be established before blanket merge clearance.

PR #7, opened 2026-07-30
feat: auto-detect AI models from provider API for both cli and web dashboard

aor-rex:
Replace the freeform model text input with an auto-detected dropdown
in both the web settings page and the CLI onboarding wizard.

**Web settings (web/src/app/settings/page.tsx):**
- New `POST /api/models` route proxies model listing requests to each
  AI provider's models endpoint (OpenAI-compatible, Anthropic, Google)
- Model field now shows a `<select>` dropdown populated from the
  provider's API when successful, with fallback to text input on error
- Debounced fetch re-triggers on provider, API key, or custom URL changes

**CLI onboarding (cli/bin.mjs):**
- Added `fetchModels()` helper that calls each provider's models API
- After provider selection + key entry, lists available models as a
  numbered picker (same style as the provider selector)
- Falls back to freeform text input when the models endpoint is unreachable

**Fixes:**
- `askSecret` masked input: uses `console.log` + silent read to ensure
  the prompt label is always visible and backspace/ANSI sequences don't
  create phantom `*` characters

### DIFF
--- a/cli/bin.mjs
+++ b/cli/bin.mjs
@@ -179,10 +179,9 @@ function makePrompter() {
   const askSecret = (q) =>
     new Promise((res) => {
       const orig = rl._writeToOutput.bind(rl);
-      process.stdout.write(q);
+      console.log(q);
       rl._writeToOutput = (s) => {
         if (s.includes("\n") || s.includes("\r")) orig(s);
-        else rl.output.write("*");
       };
       rl.question("", (answer) => {
         rl._writeToOutput = orig;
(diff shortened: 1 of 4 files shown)


kristianulv23:
Nice idea, but this seems to change something the repo already promises:

- `packages/core/src/llm-providers.ts` says OpenRouter should let you “set any model id in the model field”
- `custom` also reads like you should be able to type your own model

With this PR, that no longer works when model listing succeeds, because the text field turns into a dropdown.

I’d keep the auto-detected list, but still let people type a model manually?

aor-rex:
ouh yeah, i believe the drop down also applies to openrouter, and displaying all available model provided by openrouter. i understand what you mean, and well, i believe the dev will also review it.

PR #8, opened 2026-08-02
feat: run merrymen as a locally-built Docker image

aor-rex:
Let owners run merrymen in Docker without touching a registry — the image is built from source on their own machine, and the installers offer a choice.

Image + orchestration
- Dockerfile (repo root): node:22-bookworm-slim, cached npm ci, one-time next build; runs merrymen start (dashboard on :3100 + supervised worker). All state lives in /app/.merrymen; the container is disposable.
- .dockerignore, docker-compose.yml: same local build + bind mount of ~/.merrymen, user: ${MYUID:-1000}:${MYGID:-1000} so volume files stay host-owned, TZ + MERRYMEN_ALLOWED_HOSTS passthrough.

Installers (install.sh, install.ps1)
- Prompt at the top: 
1) Local machine (existing Node + npm path) or 
2) Docker (clone source → build the image locally → install a merrymen wrapper that drives docker). Falls back to local on no TTY; bash reads via /dev/tty so curl | bash works; git → tarball fallback when git is absent.
- On the Docker path, a server is now a one-shot self-host: the installer detects the box's public IP (hostname -I, falling back to ip -4 addr), allowlists it for the dashboard's host guard, opens port 3100 via ufw when present, starts the band, and prints the reachable http://<ip>:3100. Keys and strategy are added afterward in the dashboard at /settings.

merrymen wrapper (scripts/docker/merrymen)
- start (detached daemon, --restart unless-stopped), stop/restart/logs, update (pull + rebuild + recreate the container — a rebuilt image only reaches the daemon through a new container), and one-shot --rm runs (onboard, doctor, status, recover, …) sharing the same volume.
- Runs containers as the host user so written files are owned by you; sources ~/.merrymen-docker/env on every run, so the installer's allowlist and TZ persist across shells and reboots.

Self-hosting opt-in (web/src/middleware.ts)
- New MERRYMEN_ALLOWED_HOSTS (comma-separated, case-insensitive) lets an owner allow exact public hostnames past the dashboard's host guard — the DNS-rebind protection stays intact and is the only way a public domain passes. Loopback / private-IP hosts are unaffected. The Sec-Fetch-Site CSRF guard is untouched.

Headless fix (cli/bin.mjs)
- openBrowser used async spawn(xdg-open) whose ENOENT fires an unhandled error event — crashing merrymen start on any headless box (including this image). The error is now swallowed; the URL is still printed.

Docs — README + site install pages show the Docker option, compose usage (MYUID=$(id -u) MYGID=$(id -g)), the chown one-liner for a pre-existing root-owned ~/.merrymen, and the MERRYMEN_ALLOWED_HOSTS / LAN-only note.

Notes
- The dashboard is LAN-only by default; MERRYMEN_ALLOWED_HOSTS is the deliberate self-hosting opt-in (on a VPS the Docker installer sets it automatically to the box's IP).
- The Windows path (install.ps1, scripts/docker/merrymen.ps1/.cmd) is included but not tested by the author (no Windows env available).
- No code is changed outside the files listed; the gateway image is untouched.


millw14:
Thank you for the effort here — this is a lot of careful work, and the Dockerfile itself is good: multi-stage, `--ignore-scripts` with a stated reason, state kept on a bind mount so the container stays disposable, no npm dependencies added, and alpine rejected with an argument rather than by reflex. The UID/GID handling and the note about Docker creating a missing bind-mount source as root are the kind of detail that only comes from actually running it.

I'm closing it, for one reason of substance and two defects.

**The substance: this changes what merrymen is.** The dashboard has no login and can move funds. Today that is safe because it is bound to loopback on a machine the owner controls, and `web/src/middleware.ts` rejects non-loopback `Host` headers and cross-site `Sec-Fetch-Site` — a guard that exists because an audit found a CSRF that could sweep the whole account. This PR publishes that dashboard with `-p 3100:3100` and `MERRYMEN_HOST=0.0.0.0`, which binds every host interface: the LAN on a home box, the open internet on a VPS, over plain HTTP. The installer then detects the public IP and writes it into `MERRYMEN_ALLOWED_HOSTS` specifically so it passes the host guard.

The description says "the DNS-rebind protection stays intact… the Sec-Fetch-Site CSRF guard is untouched". The code is untouched; its protection is not. Allowlisting the public IP is what that guard exists to prevent, and `Sec-Fetch-Site` does nothing against someone typing the URL directly.

Remote access isn't unreasonable to want. But it needs authentication designed for it — not an allowlist entry — and that is a product decision rather than a packaging one. Merging this would ship "your keys, your caps, on your machine" while quietly making the control surface reachable from the internet.

**Defect 1 — arbitrary code execution.** The installer does:

```sh
SERVER_IP="$(curl -fsS4 --max-time 5 https://api.ipify.org 2>/dev/null || true)"
printf 'MERRYMEN_ALLOWED_HOSTS=%s\n' "$SERVER_IP" > "$CFG"
```

and the README says that file is "sourced on every run" by the `merrymen` wrapper. An unvalidated third-party HTTP response, written to a file a shell `source`s, on the machine holding the owner key. Anyone who can answer for `api.ipify.org` — the service itself, a MITM on plain HTTP, a DNS hijack, a hostile network — gets shell. Even keeping the feature, that value needs validating as an IP literal before it is written, and ideally never `source`d.

**Defect 2 — a mutable base image.** `node:22-bookworm-slim` is a moving tag. For a wallet, that should be digest-pinned so the same commit produces the same image.

If you'd like to take another run at it, the piece I'd genuinely welcome is **the Dockerfile and compose file alone, loopback-only** — `-p 127.0.0.1:3100:3100`, no ipify, no ufw, no host allowlisting. That is a real convenience for people who would rather not install Node, and it changes no security property. The remote-hosting story should come after there's an auth story to attach it to.

aor-rex:
thanks for the detailed review on this pr. i understand the decision on the loopback boundary. i'll resubmit as just dockerfile + compose, loopback only (-p 127.0.0.1:3100:3100) with no ipify, ufw, or host allowlisting. i'll digest-pin node:22-bookworm-slim too so the same commit always builds the same image. the installer still lets people pick local vs docker and builds from source; keeps the headless openBrowser fix. no security property changes.

on remote hosting, since there's no auth gate yet, here's an alternative that doesn't need one: a private vpn (wireguard/tailscale). it's not public exposure as it extends the owner's trusted network the same way the LAN already does, and the host guard already passes private ranges, so only clients on that vpn can reach the dashboard. happy to open an issue or a pr for it separately, whenever you'd like to look.

the auth-gated public exposure i'll leave out entirely. that's a product call, and happy to open an issue instead if that's out of scope.

PR #10, opened 2026-08-06
fix: surface the full command list in the Telegram bot menu

aor-rex:
## Problem

The Telegram bot used Telegram's bot API but never registered its commands, so
the "/" button in chat showed an **empty menu** — users had to know commands by
heart or via `/help`. The full `setMyCommands` path was simply missing.

## Change

- `worker/src/telegram/api.ts` — add `BOT_COMMANDS`: 47 entries mirroring the
  closed slash-command set (`parseSlash`), as Telegram-valid `BotCommand`
  objects (3-32 lowercase chars, plain-text descriptions, no HTML). The two
  2-char commands `/ls` and `/pc` work in chat but can't be menu entries, so
  they're intentionally omitted. Plus a best-effort `setMyCommands()` client
  that never throws.

- `worker/src/telegram/service.ts` — register the menu **once per token** at
  the top of the poll loop, so it fires on enable-after-start, re-registers on
  token change, and never spams. Silent on success; logs a one-time warning on
  failure without breaking the poll.

## Tests

`npm test` — 695 pass, 0 fail (added `BOT_COMMANDS` validity + `setMyCommands`
request/degrade cases). `npm run typecheck` clean.

Edit: Added `/ls` and `/pc` to command menu as they are also telegram-valid `BotCommand`

### DIFF
--- a/worker/src/telegram/api.test.ts
+++ b/worker/src/telegram/api.test.ts
@@ -148,8 +148,8 @@ describe("BOT_COMMANDS — the Telegram command menu", () => {
     assert.ok(BOT_COMMANDS.length > 0);
     const seen = new Set<string>();
     for (const { command, description } of BOT_COMMANDS) {
-      // command: 3-32 lowercase alnum/underscore, no leading slash, unique
-      assert.match(command, /^[a-z][a-z0-9_]{2,31}$/);
+      // command: 1-32 lowercase alnum/underscore, no leading slash, unique
+      assert.match(command, /^[a-z][a-z0-9_]{0,31}$/);
       assert.ok(!command.startsWith("/"));
       assert.equal(seen.has(command), false, `duplicate command /${command}`);
       seen.add(command);
--- a/worker/src/telegram/api.test.ts
+++ b/worker/src/telegram/api.test.ts
@@ -1,6 +1,7 @@
 import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
 import { describe, it } from "node:test";
-import { BOT_COMMANDS, esc, getMe, getUpdates, sendMessage, setMyCommands, type FetchLike } from "./api";
+import { BOT_COMMANDS, esc, getMe, getUpdates, sendMessage, setMyCommands, publicBotCommands, type FetchLike } from "./api";
 import { parseSlash } from "./interpreter";
 
 /** Fake fetch capturing the last call, returning a canned envelope. */
@@ -186,6 +187,22 @@ describe("setMyCommands", () => {
     assert.deepEqual(body.scope, { type: "all_private_chats" });
   });
 
+  it("serializes a per-chat scope (full menu pushed to one allowlisted chat)", async () => {
+    const f = fakeFetch(200, OK(true));
+    await setMyCommands({ token: "123:abc", fetchFn: f }, undefined, { type: "chat", chat_id: -100111 });
+    const body = JSON.parse(f.lastBody!) as { commands?: unknown; scope?: { type: string; chat_id?: number } };
+    assert.deepEqual(body.commands, BOT_COMMANDS); // no list given → the full owner menu
+    assert.deepEqual(body.scope, { type: "chat", chat_id: -100111 });
+  });
+
+  it("pushes exactly publicBotCommands when told (the stranger menu)", async () => {
+    const f = fakeFetch(200, OK(true));
+    await setMyCommands({ token: "123:abc", fetchFn: f }, publicBotCommands, { type: "all_private_chats" });
+    const body = JSON.parse(f.lastBody!) as { commands: { command: string }[] };
+    assert.deepEqual(body.commands, publicBotCommands);
+    assert.ok(!body.commands.some((c) => c.command === "run"));
+  });
+
   it("degrades gracefully on ok:false (bad token or unsupported)", async () => {
     const f = fakeFetch(200, { ok: false, description: "Unauthorized" });
     const { ok, reason } = await setMyCommands({ token: "bad", fetchFn: f });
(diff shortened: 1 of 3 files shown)


millw14:
Good, small, and it does what it says. I checked the thing I check first on any Telegram PR — whether a "menu" change quietly adds a route into `executeCommand`, which is why #3 was closed — and **it does not**. The service change is one import plus a 7-line block inside `pollOnce()`, sitting before `getUpdates` and outside `handle()` entirely. It touches neither allowlist, neither gate, nor `parseSlash`, the `Command` union, `CONTROL_KINDS`, `PC_KINDS` or `executor.ts`. `BOT_COMMANDS` is a plain const whose only consumers are `setMyCommands` and the new test. `/ls` and `/pc` both remain behind the sender gate and the capability check.

One change I'd want before merging, and two worth considering.

### Please scope the menu to private chats

`setMyCommands` is called with no `scope`, so Telegram applies `BotCommandScopeDefault` — "all users in private chats **and all other chat members**". Every member of any group the bot sits in now gets a 50-entry menu advertising `/shutdown`, `/run`, `/type`, `/key`, `/clip`, `/lock`, `/sleep`, `/agent`, `/transfer`, `/buy`, `/sell` and `/cap`. And because `BOT_COMMANDS` is static, it advertises all of them on a default install where PC control is **off**.

Nothing is bypassed — I confirmed the gates hold. But merrymen's posture on PC control is quiet-by-default, and this broadcasts it. One line fixes it:

```js
await setMyCommands({ token: cfg.telegramBotToken }, undefined, { type: "all_private_chats" });
```

If you want a group menu, register a read-only subset (`status`, `positions`, `pnl`, `trades`, `report`, `why`, `brag`) under `all_group_chats`.

### While you're there — `/forget` is now a button for people the gate doesn't cover

The sender gate at `service.ts:605-611` covers `CONTROL_KINDS`, `PC_KINDS`, `transfer`, `confirm` and `agent`. It does **not** cover `forget`, `name`, `remember`, `cancel`, `alert`, `unalert`, `remind`, `unremind`. A non-allowlisted member of an allowlisted group can already run `/forget` — which wipes owner memory and chat history — but until now they had to know it existed. This PR makes it a labelled button in their menu.

The gap is pre-existing and not yours to fix, but scoping the menu removes the reachability this PR adds. Worth closing properly at some point by inverting that set into an explicit `READ_ONLY_KINDS` allowlist, so a new kind defaults to gated rather than open.

### `/kill` is missing from "the full command list"

`HELP_TEXT` has `/kill — destroy the grant, stand the band down` and it's in `CONTROL_KINDS`, but there's no `kill` entry in `BOT_COMMANDS` — so the one command someone reaches for in an emergency is the one absent from the menu, while the equally destructive `/shutdown` is present. Conversely `/sleep` and `/shutdown` appear nowhere in `/help`.

That's the cost of two hand-maintained lists with nothing tying them together. The new test asserts Telegram *formatting* validity but never calls `parseSlash`, so a menu entry for a command that doesn't exist would ship green. A test that runs every `BOT_COMMANDS` entry through `parseSlash` and asserts it isn't the unknown-command branch would catch that class permanently — allowlisting `agent`, which is routed before `parseSlash`.

Then decide on `/kill` deliberately: add it for parity, or leave it out with a comment saying a one-tap misfire shouldn't destroy the grant. Either is defensible; silence isn't.

Happy to merge once the scope is set.

aor-rex:
thanks for the review.

`setMyCommands` is now scoped to `all_private_chats`. added the `/kill` entry for parity with `/help` and the gate, since scoping the menu means only the owner sees it.

also added the cross-check test you suggested  , every `BOT_COMMANDS` entry runs through`parseSlash` (skipping `/agent`, which routes before it) and fails if it hits the unknown branch, so the menu can't drift from the parser again.

on the `/forget` gap, agreed it's pre-existing, and scoping the menu removes the reachability this pr added. leaving the `READ_ONLY_KINDS` inversion for a follow-up.

millw14 (commented):
Thanks — this is a real papercut, and the implementation is the right shape: pushed once per token, best-effort, and a failure can never break the poll loop. I particularly like the test that runs every menu entry through `parseSlash`, so the menu can't advertise a command that doesn't exist.

Two things to fix:

1. **The test only guards one direction, and the menu is already stale.** It asserts every *menu entry* is a real command, but nothing asserts every *real command* is in the menu. `/depth` landed on `main` after you opened this and isn't listed; nor is `/digest`. Since silent drift is the exact bug this PR exists to fix, please add the reverse assertion — the top-level commands `parseSlash` handles should all appear in `BOT_COMMANDS`, with an explicit allowlist for the aliases you deliberately don't surface.

2. **The "once per token" flag is set before the await:**

   ```ts
   commandsRegisteredFor = cfg.telegramBotToken;
   const menu = await setMyCommands(...);
   ```

   One transient network failure at startup and the menu is lost until a restart or a token change. Assign after a successful call instead.

One thing I'd like your opinion on rather than a change request: the menu is pushed with scope `all_private_chats`, so any stranger who messages the bot sees the full PC-control surface — `/run`, `/type`, `/shutdown`, `/agent`. Every one of them is gated at runtime so nothing is actually exposed, but it does advertise the remote-control surface to un-allowlisted users. Scoping the push to allowlisted chats, or splitting the PC commands into a second scoped push, would fit the project's posture better. Happy to hear the argument for keeping it simple.

Heads up on ordering: this touches `worker/src/telegram/api.ts`, as do #11 and #13. I plan to land this one first, so those two will need a rebase.

millw14:
To land this, when you get a chance:

1. The two fixes from my review — the reverse drift assertion, and moving the `commandsRegisteredFor` assignment after a successful `setMyCommands` call.
2. Rebase onto current `main`. This branch is about three weeks behind and `main` has since picked up `/depth` and `/digest`, which is also what the reverse assertion will catch.

```bash
git fetch upstream
git rebase upstream/main
git push --force-with-lease
```

Pushing to the same branch updates this PR in place — no need to open a new one.

This is the first of the three Telegram PRs I'm landing, so #11 and #13 will rebase onto it rather than the other way around.

aor-rex:
Both review fixes are in, rebased onto current `main` (`b525450`).

### 1. Reverse drift guard
The menu → `parseSlash` test now also checks `parseSlash` → menu. Every `case` label in `parseSlash` (`interpreter.ts`) is scraped from source and must be in `BOT_COMMANDS` or an explicit `HIDDEN_ALIASES` allowlist (32 deliberate synonyms — `send`/`withdraw` for `transfer`, `liquidity`/`levels` for `depth`, etc.). The list itself can't rot, and each alias must map to a canonical entry that *is* advertised. This is what caught `/depth` going missing — **surfaced now**; `/digest` stays a deliberate alias of `report`.

### 2. Register-after-success
`commandsRegisteredFor` moved **behind** the `await setMyCommands` and is keyed on `token + sorted allowlist`. One transient failure at startup retries next poll instead of losing the menu until restart; warns once per config.

### 3. Scoping — went with your split
- **Strangers** → `publicBotCommands` (14 reads/signposts) via `all_private_chats`: `help`, `status`, `positions`, `depth`, `pnl`, `trades`, `report`, `why`, `brag`, `alerts`, `reminders`, `soul`, `wallet`, `link`
- **Owner** → full `BOT_COMMANDS` via `chat` per allowlisted id — keeps `/run`, `/type`, `/agent` discoverable where it matters

Not showing them would hide that they exist (`/help` lists everything, but the `/` menu is how people discover). `publicBotCommands` is pinned so it can't quietly grow.

> **Note:** members of allowed groups **cannot use** PC commands — every state-changing / PC command is gated per-sender (`allowlist.includes(fromId)`), so a stranger in an allowlisted group sees the entry but is refused at runtime. Nothing is exposed.

Rebased, `948/948` green, `typecheck` clean.

---

**also a question:** would you prefer the bot be **owner-scoped entirely** on Telegram — i.e. only the owner can use/interact with the bot at all (not just PC control), with all menus and all commands gated to allowlisted chats / `chat_member`? Happy to make that the posture if you want it.

PR #11, opened 2026-08-06
feat: telegram link-code onboarding modal + inline confirm/cancel for parked actions

aor-rex:
## Problem

Two rough edges in Telegram onboarding/control:
- A fresh install shows the dashboard's "link this bot" CTA but the one-time
  `/link <code>` handshake lived buried in settings — users who landed on the
  dashboard had no guided path to the code past the initial connect.
- Parked actions (`/transfer`, `/shell`, `/getfile`, ...) could only be
  resolved by typing `/confirm` or `/cancel`. The chat model answered a
  plain-language "yes, do it" from /blind/ — it never learned an action was
  pending, because slash-parked replies aren't in history and state had no
  marker.

## Change

**1. Link-code onboarding modal (dashboard)**

- `web/src/components/TelegramLinkModal.tsx` (new) — global modal, polls
  `/api/telegram` every 5s, shows once a token is connected but the bot is
  unclaimed. Three states: code ready (copy button + "Open @bot" link),
  preparing, and turn-on (`/settings#telegram`).
- Auto-closes the polling cycle the bot gets claimed; dismiss (X /
  click-outside / Escape) is not persisted, so it re-shows until linked.
- Wired into `web/src/app/layout.tsx`; styling in `web/src/app/globals.css`.

- **2. Inline Confirm/Cancel buttons** (`worker/src/telegram/api.ts`,
  `worker/src/telegram/service.ts`)
  - `sendMessage` gains `reply_markup`; new `editMessageText()` + 
    `answerCallbackQuery()`; `getUpdates` now subscribes to `callback_query`.
  - Any reply that just parks an action attaches a ✅ Confirm / ✖ Cancel
    keyboard. Taps resolve through the *same* gated confirm/cancel path as
    typing `/confirm`, keyed to the parking user.
  - Deps extracted to `buildCmdDeps(cfg, peer)` so book and buttons share one
    route; planner `describePending()` names the parked action.

## Screenshots

**1. Link-code onboarding modal** — one-time code + copy + "Open @bot"
photo_2026-08-06_13-53-09

**2. Inline Confirm / Cancel on a parked action**
photo_2026-08-06_13-53-07

### DIFF
--- a/web/src/app/globals.css
+++ b/web/src/app/globals.css
@@ -569,6 +569,54 @@ button { font: inherit; cursor: pointer; }
 }
 .tg-cta-btn:hover { background: #1c93d0; }
 
+/* ---------- telegram link modal (global first-link onboarding) ---------- */
+.tg-modal-backdrop {
+  position: fixed; inset: 0; z-index: 100;
+  display: flex; align-items: center; justify-content: center;
+  padding: 20px;
+  background: color-mix(in srgb, var(--bg) 62%, transparent);
+  backdrop-filter: blur(3px);
+  animation: tg-fade 160ms ease-out;
+}
+.tg-modal {
+  position: relative;
+  width: min(380px, 100%);
+  display: flex; flex-direction: column; gap: 12px;
+  padding: 22px 20px 20px;
+  background: var(--bg-2);
+  border: 1px solid var(--border-2);
+  border-radius: var(--radius);
+  box-shadow: 0 18px 50px rgba(0, 0, 0, 0.5);
+  animation: tg-pop 180ms ease-out;
+}
+.tg-modal-close {
+  position: absolute; top: 10px; right: 12px;
+  width: 28px; height: 28px; line-height: 1;
+  border: none; background: none; color: var(--text-faint);
+  font-size: 22px; cursor: pointer; border-radius: 6px;
+}
+.tg-modal-close:hover { color: var(--text); background: var(--bg-3); }
+.tg-code-row { display: flex; align-items: stretch; gap: 8px; }
+.tg-code {
+  flex: 1;
+  font-family: var(--mono); font-size: 17px; font-weight: 700;
+  letter-spacing: 4px; text-align: center;
+  color: var(--green-soft);
+  background: var(--bg-3); border: 1px solid var(--border-2);
+  border-radius: 8px; padding: 10px 8px;
+  user-select: all;
+}
+.tg-copy-btn {
+  font-family: var(--mono); font-size: 12px; font-weight: 700;
+  color: var(--text); background: var(--bg-3);
+  border: 1px solid var(--border-2); border-radius: 8px;
+  padding: 0 12px; cursor: pointer; white-space: nowrap;
+}
+.tg-copy-btn:hover { border-color: var(--green-deep); color: var(--green-soft); }
+.tg-modal-note { font-size: 10.5px; color: var(--text-faint); }
+@keyframes tg-fade { from { opacity: 0; } to { opacity: 1; } }
+@keyframes tg-pop { from { opacity: 0; transform: translateY(8px) scale(0.98); } to { opacity: 1; transform: none; } }
+
 /* ---------- the wall (trust layer) panel ---------- */
 .wall-panel { display: flex; flex-direction: column; gap: 10px; }
 .wall-sub { font-size: 11.5px; color: var(--text-dim); line-height: 1.55; }
--- a/web/src/app/layout.tsx
+++ b/web/src/app/layout.tsx
@@ -2,6 +2,7 @@ import type { Metadata, Viewport } from "next";
 import { Hanken_Grotesk, JetBrains_Mono } from "next/font/google";
 import "./globals.css";
 import { RegisterSW } from "@/components/RegisterSW";
+import { TelegramLinkModal } from "@/components/TelegramLinkModal";
 
 // The merrymen.dev typefaces — used on the setup/settings screens (.setup-look)
 // so onboarding feels like the website; the trading terminal keeps its own fonts.
@@ -50,6 +51,7 @@ export default function RootLayout({ children }: { children: React.ReactNode }) 
       <body>
         {children}
         <RegisterSW />
+        <TelegramLinkModal />
       </body>
     </html>
   );
--- a/web/src/components/TelegramLinkModal.tsx
+++ b/web/src/components/TelegramLinkModal.tsx
@@ -1,0 +1,134 @@
+"use client";
+
+import Link from "next/link";
+import { useEffect, useState } from "react";
+import type { TelegramStatus } from "@/app/api/telegram/route";
+
+/**
+ * Global first-link onboarding popup. Appears on any dashboard/settings page as
+ * soon as the user has a valid Telegram token AND the bot is reachable, but has
+ * not linked an owner chat yet. It guides them to the one-time /link code and
+ * auto-closes the moment the worker confirms a link (ownerId appears).
+ *
+ * Deliberately NOT persisted: dismissal lasts only for the lifetime of the open
+ * page. A fresh visit re-shows the popup until the bot is actually linked, so
+ * the "I can't see the code" gap can't be dismissed away forever.
+ */
+export function TelegramLinkModal() {
+  const [tg, setTg] = useState<TelegramStatus | null>(null);
+  const [dismissed, setDismissed] = useState(false);
+  const [copied, setCopied] = useState(false);
+
+  useEffect(() => {
+    let alive = true;
+    const load = async () => {
+      try {
+        const r = await fetch("/api/telegram");
+        if (alive && r.ok) setTg((await r.json()) as TelegramStatus);
+      } catch {
+        /* keep last */
+      }
+    };
+    load();
+    const id = setInterval(load, 5000);
+    return () => {
+      alive = false;
+      clearInterval(id);
+    };
+  }, []);
+
+  const open = !!tg && tg.hasToken && tg.connected && tg.ownerId === null && !dismissed;
+
+  useEffect(() => {
+    if (!open) return;
+    const onKey = (e: KeyboardEvent) => {
+      if (e.key === "Escape") setDismissed(true);
+    };
+    window.addEventListener("keydown", onKey);
+    return () => window.removeEventListener("keydown", onKey);
+  }, [open]);
+
+  if (!open || !tg) return null;
+
+  const botLink = tg.botUsername ? `https://t.me/${tg.botUsername}` : null;
+  const code = tg.linkCode;
+  const preparing = !!tg.enabled && !code;
+  const needsTurnOn = !tg.enabled && !code;
+
+  const copyCode = async () => {
+    if (!code) return;
+    try {
+      await navigator.clipboard.writeText(code);
+    } catch {
+      const ta = document.createElement("textarea");
+      ta.value = code;
+      ta.style.position = "fixed";
+      ta.style.opacity = "0";
+      document.body.appendChild(ta);
+      ta.select();
+      document.execCommand("copy");
+      document.body.removeChild(ta);
+    }
+    setCopied(true);
+    setTimeout(() => setCopied(false), 1500);
+  };
+
+  return (
+    <div
+      className="tg-modal-backdrop"
+      onClick={(e) => {
+        if (e.target === e.currentTarget) setDismissed(true);
+      }}
+    >
+      <div className="tg-modal" role="dialog" aria-modal="true" aria-label="Link your Telegram bot">
+        <button type="button" className="tg-modal-close" aria-label="Dismiss" onClick={() => setDismissed(true)}>
+          ×
+        </button>
+        <div className="tg-card-title">💬 Link your Telegram</div>
+
+        {code ? (
+          <>
+            <p className="tg-card-sub">
+              Open <b>{tg.botUsername ? `@${tg.botUsername}` : "your bot"}</b> and send{" "}
+              <code>/link {code}</code> — the first message that claims it makes you its owner.
+            </p>
+            <div className="tg-code-row">
+              <code className="tg-code">{code}</code>
+              <button type="button" className="tg-copy-btn" onClick={copyCode}>
+                {copied ? "Copied ✓" : "Copy"}
+              </button>
+            </div>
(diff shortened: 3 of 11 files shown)


millw14:
Thanks for this — the link-code modal is a genuine fix for a real papercut, and the refactor underneath is careful work: hoisting `CommandDeps` and the `/link` closure into peer-parameterised `buildCmdDeps(cfg, peer)` so both paths build identical deps is the right shape, and I diffed that move field by field — it's faithful, the only deltas are `msg.*` → `peer.*`. Removing `confirm` from the LLM's choosable kinds and replacing it with "there is no confirm/cancel kind for you to choose" is also exactly right.

Four things need fixing before it can land. The first is the one I'd fix even if nothing else changed.

### 1 · Blocker — the Confirm button isn't bound to the action it's attached to

`callback_data` is the bare literal `"confirm"` (service.ts:117-123), and `handleCallback` resolves it against whatever occupies the single pending slot *at tap time* — `cb.messageId` is never consulted (service.ts:703-723). Three details combine into a real problem:

- `executeCommand` overwrites the pending slot unconditionally — `case "transfer"` and the five PC park kinds all call `setPending(...)` with no check for an existing one.
- `const parked = !hadPending && pending.has(...)` means the **second** park deliberately gets no buttons — so the only tappable keyboard left in the chat is the superseded one.

Sequence: `/transfer 0xAAA 5` renders "⚠️ confirm transfer — send 5 USDG to 0xAAA" with buttons. Then `/shell <cmd>` parks over it and arrives with no buttons. Scroll up, tap ✅ Confirm on the first message — **the shell command runs.** The button affirmatively misrepresents what is being approved, on the two paths that matter most.

And the second park needn't be a slash command: the LLM can route plain English into any parked kind, so model output can change what an already-displayed Confirm button will execute. The verbatim-address defence in `coerceLlmCommand` still stops a model substituting an *address* — it doesn't stop it substituting the *action* behind a live button.

**Fix:** put a short nonce on `PendingAction` (it's already an expiring record), emit `callback_data: \`confirm:${nonce}\`` / `\`cancel:${nonce}\`` — well inside Telegram's 64-byte cap — and in `handleCallback` parse `[verb, nonce]` from a closed set, refusing when it doesn't match and editing the stale message to "this ask was superseded". Separately drop the `!hadPending` condition so the newest park always carries its own keyboard, and have `executeCommand` refuse to park over a live pending rather than silently replacing it.

### 2 · High — `handleCallback` applies no authorization at all

It never consults `cfg.telegramAllowlist`. The message path checks twice — the chat-level `allowed` at service.ts:446 and the sender-level group gate at 653-666, which names `cmd.kind === "confirm"` explicitly. The callback path checks neither; the only thing between a tap and `executeCommand` is `pending.has(\`${chatId}:${fromId}\`)`.

I want to be accurate about how bad this is, because my reviewers split on it. It is **not** trivially exploitable today: `pending` can only be populated through the gated message path, so an un-allowlisted stranger has no slot to confirm. But that's a *derived* property, not an enforced one — and it breaks on the documented revocation move. Allowlist a member, they park a transfer, you notice and remove them from `telegramAllowlist`, they tap Confirm: typing `/confirm` is rejected at 653-666, tapping is not. The funds leave.

The doc comment above the function claims taps run "the exact same executor confirm/cancel branch as typing /confirm (same re-vetting, same gates)". The executor's gates are **capability** gates (`transferEnabled`, `pcRefusal`); authorization lives in `handle`, which `handleCallback` doesn't call. Worth correcting either way.

**Fix:** re-apply both gates at the top of `handleCallback` against `cb.fromId`, before the `pending.has` check so revocation revokes buttons too.

### 3 · High — any group member can wipe the owner's confirm

The not-yours branch (service.ts:715-719) runs with no authorization and calls `editMessageText` on the **owner's** message. So any member of a chat the bot is in can tap the owner's Confirm button and rewrite it to "nothing pending to confirm — the ask has expired or already resolved." They can't execute it — the pending key protects that — but they can make the owner believe their transfer already resolved. Only edit a message when the tapper owns the pending.

### 4 · High — expired pendings are never swept, so buttons silently stop appearing

`hadPending` is computed from `pending.has(...)`, and nothing removes an entry on TTL expiry. Once a park expires without being resolved, the key stays, `hadPending` is permanently true for that chat+user, and every future parked action arrives with **no buttons at all** — the feature quietly dies for that user with no error. Sweep on read, or check `expiresAt` in `hadPending`.

---

Smaller: the `PENDING CONFIRM:` hint that `describePending()` feeds into LLM state is new — worth a sentence in the PR body on why the model needs to know, given the same PR removes its ability to act on it.

None of this is structural — the architecture is right, the gates just need to exist on the new path. Happy to look again quickly once these are in.

aor-rex:
i have addressed all four of your review point:

- the **nonce binding**. Buttons emit `confirm:<nonce>` / `cancel:<nonce>` matched against the live pending, so a stale button can't confirm a newer ask. `executeCommand` also refuses to park over a live pending instead of silently replacing it.
- **callback authorization**. `handleCallback` re-runs both message-path gates (chat allowlist + group sender rule) against the tapper, revocation kills buttons instantly.
- **owner-only edits**. A message is only edited when the tapper owns the pending slot; strangers/expired asks get a toast, never an edit.
- **TTL sweep**. Expired pendings are swept on read, so a lapsed slot stops blocking new parks and stops steering the model's hint.

Extracted the resolution logic into `telegram/callback.ts` (injected deps) and added unit tests covering all four behaviors plus failure/fallback paths.

millw14:
Heads-up: this branch now conflicts with `main`, and the conflict is my doing, not yours.

[`eaeb75a`](https://github.com/millw14/merrymen/commit/eaeb75a) added a `/depth` read command, which touched the two spots in `worker/src/telegram/service.ts` your branch also edits:

- `TelegramServiceDeps` gained a `readDepth: (symbol: string) => Promise<string>;` member,
- `cmdDeps.reads` gained a `depth: (symbol: string) => deps.readDepth(symbol),` line.

The resolution is additive — keep both sides. Your changes and mine don't overlap semantically; they just landed on adjacent lines of the same interface and the same `reads` block. (`executor.ts`, `interpreter.ts` and `interpreter.test.ts` auto-merge cleanly; `service.ts` is the only file needing a hand.)

A plain `git merge origin/main` (or rebase, your preference) with those two hunks kept both-sides should do it. CI on `main` is green with the depth command in, so once the conflict clears the required checks should pass as they did on your last run. Sorry for the churn — the review itself stands as posted.

millw14 (commented):
I went into the callback path looking for the usual holes and you closed them. Worth recording what's right, because it's most of the PR:

- The pending slot stays keyed `chat:from`, so only the user who parked an action can resolve it.
- A tap runs through **the same `executeCommand` confirm branch** as a typed `/confirm` — same re-vetting, same gates. There is no second, weaker confirmation route, which is the thing I'd have rejected this over.
- The gates match the message path exactly, including the group rule (`service.ts:173` and `:613`).
- A tap that isn't yours gets a toast and **never edits a message**; the nonce refuses a superseded button instead of confirming whatever landed in the slot later.

One thing I need changed before this can merge:

**`describePending` sends parked-action payloads to the LLM provider.** `pendingHint` goes into `routeCtx.state`, and `describePending` renders the `type` text and `transfer <amount> USDG -> <address>` verbatim. So a user who runs `/type <a password>` and then chats has just sent that password to Groq/Anthropic/whatever custom endpoint is configured. On `main` a `/type` payload never leaves the machine — this is a new egress path, and it cuts against the redaction work in 0.10.3.

The hint's job is only to let the model say "tap the confirm button", so describing by **kind** is enough: "a keyboard action is waiting", "a transfer is waiting". No payload needed.

Two smaller things:

- **A behavior change is buried in the diff:** a live pending now *blocks* a new park ("an action is already waiting") where the documented behavior was "the latest ask wins". I think you're right that this is what makes a stale button safe — but it deserves a line in the PR description, and the refusal should say *what* is waiting. `describePending` (kind only, per above) is right there.
- **The same four-line guard is copy-pasted six times** across the transfer/shell/getfile/type/hotkey/power branches. One `blockIfPending(deps)` helper.

Nit: `cb.data.split(":", 2)` truncates rather than capturing the remainder. Harmless while nonces are hex, but it will bite silently if the format ever grows a colon.

Last thing — this bundles a fairly large `service.ts` refactor (extracting `buildCmdDeps(cfg, peer)`, +307/-214) with the feature. The refactor is good and I think it's genuinely load-bearing for the callback path, so I'm not asking you to split it, but it's most of why this now conflicts with `main`. Please rebase; #10 is landing first.

millw14:
Two things to get this landed:

1. The `describePending` change from my review — describe the parked action by **kind** only, so `/type` payloads and transfer addresses stop being sent to the LLM provider. That's the one I'm actually holding this on.
2. Rebase onto `main`, which this now conflicts with. I'm landing #10 first, and it touches `worker/src/telegram/api.ts` too, so rebase after that goes in.

```bash
git fetch upstream
git rebase upstream/main
git push --force-with-lease
```

Pushing to the same branch updates this PR in place.

The `blockIfPending` helper and the `split(":", 2)` nit are nice-to-haves — don't hold the PR for them if you'd rather I take them as follow-ups.

aor-rex:
**`describePending` no longer leaks** — `executor.ts:42` now returns kind-only (`a transfer / a keyboard action is waiting…`) — a pasted password in `/type` stays on the card (`esc()`) and never goes to the LLM. The `PENDING CONFIRM:` hint (`service.ts:599`) only steers to `✅ Confirm`. Tests at `interpreter.test.ts:639` assert no `s3cr3t` leak.

**Behavior change noted:** Live pending now blocks a new park (`a transfer is waiting… — /confirm…`) where prior documented behavior was `latest ask wins` — intentional to make stale buttons safe.

millw14:
**Needs a rebase, and it now sits behind two other Telegram PRs.**

CONFLICTING against current `main`. The worker-side files it changes (`telegram/api.ts`, `executor.ts`, `interpreter.ts`, `service.ts`) all still exist, so this is a rebase rather than a rewrite. Two things to know before starting:

1. **One file is gone.** `web/src/app/globals.css` was deleted by the terminal rewrite (#96, #103), so the modal's styling needs re-homing into the current terminal CSS rather than merged back.
2. **Order matters.** #10 (full command list in the bot menu) merged today and changed `telegram/api.ts` and `telegram/service.ts`. #107 is still open and changes `interpreter.ts`, `executor.ts` and `service.ts`. Rebasing this before those settle will mean doing the same conflicts twice.

**What to do:** rebase after #107 is resolved, and confirm the link-code flow is not already covered by the guided onboarding that shipped in #96. If it is covered, close this instead. The inline confirm/cancel for parked actions is the part most likely to still be worth having.

---

aor-rex:
Rebuilt per the sequencing ask: the old tip (August `main`, 648 behind, carrying files #13 now covers) is replaced by a narrow extraction cherry-picked onto current `main`, pushed over this same branch — no new PR. What's in the two rebuilt commits: the link-code onboarding modal (`TelegramLinkModal.tsx` + root-layout mount) and nothing else it shouldn't. Verified redundant against current `main` and deliberately left out: the steering prompt (keys on a `PENDING CONFIRM` marker nothing emits — the injection lived in the dropped button commit), the permission-message fixes (all four hunks already superseded on `main`), and the button/markup half (lives in #13). Deliberately deferred: the `callback.ts` module + tap-surfacing — it requires nonced tap payloads while #13's green design uses bare `confirm`/`cancel`, so that migration belongs after #13's format settles, not as a competing design here. One fix found during the rebuild: the modal styles had auto-merged into quarantined (unloaded) `legacy.css` — moved to loaded `terminal/forms.css`, without which the modal renders unstyled. Local: `npm ci` + typecheck (worker/web/browser) + full suite (4988 pass / 0 fail). CI on the rebased tip: `app` and `contracts` green.

millw14 (commented):

**[P2] The modal does not establish keyboard focus or modal containment** — `web/src/components/TelegramLinkModal.tsx:83`. The overlay declares `aria-modal="true"`, but opening it leaves focus wherever it was in the underlying dashboard, and Tab can continue to controls behind the overlay. Keyboard users can operate obscured controls instead of the link-code dialog. Move focus into the dialog on open, contain Tab/Shift+Tab, make the background inert, and restore focus on dismissal; a native dialog or the shared dialog primitive can provide this. Verify with keyboard-only navigation.

Validation: source review of polling, dismissal, copy handling and global placement, plus existing green app/contracts/gateway checks. No interactive browser/accessibility test was run. This diff implements the link modal only; the title's inline confirm/cancel feature is not implemented in these three changed files.

aor-rex:
Confirmed and fixed — thank you. The overlay declared `aria-modal" while Tab reached the dashboard behind it.\n\n**Fix:** focus moves into the dialog on open, Tab/Shift+Tab cycles buttons AND links (the modal has CTA anchors, not just buttons), stray focus is pulled back, Escape dismisses, previous focus restores. Extracted as shared `useFocusTrap` — same behavior as FirstVisit's tour card, so the next dialog doesn't reinvent it a third time; the modal consumes the hook.\n\n**Evidence (the keyboard-only verification you asked for):** new jsdom tests drive synthetic Tab/Shift+Tab/Escape through the hook and assert enter-inside, full cycle incl. links, pull-back, and dismiss — 3/3 green. Full suite green locally; `app`+`contracts`+`gateway` green on the pushed tip (app needed one re-run: an unrelated Gecko pacing-cache test flaked under load, passes alone 5/5, zero shared code with this change). No interactive browser run claimed — stated plainly per your standard.

millw14 (commented):

**Needs changes.** Initial focus, keyboard containment and restoration are now implemented. One interaction between the hook and its caller undermines keyboard use.

**[P2] Keep the focus-trap effect stable across ordinary renders** — `web/src/components/use-focus-trap.ts` (effect dependency `onDismiss`) and `TelegramLinkModal.tsx:49`. The modal passes a new inline dismissal callback every render. Each 5-second Telegram poll updates `tg`, which tears down the effect, restores background focus, and reruns it to focus the dialog root. Copy-state changes do the same. A keyboard user on Copy or the Telegram link unexpectedly loses their place without closing the dialog. Memoize the callback or keep the latest callback in a ref without reinstalling the trap. Add a rerender/poll test asserting focus remains on the current control.

Verification: source and supplied test inspection, not an interactive accessibility run or local execution of contributor code. The PR also conflicts with current main and needs updated CI. No automatic merge authorization.

PR #12, opened 2026-08-06
feat: first-run onboarding — dashboard wizard + terminal channel choice

aor-rex:
First-run setup on the web, with the terminal kept as a first-class choice.

## Dashboard wizard

- `web/src/components/OnboardWizard.tsx` (new) — a full-screen 5-step overlay on `/`, shown only on
  brand-new installs (`!webOnboarded && nothing meaningful configured`). Mirrors `merrymen onboard`:
  welcome → AI provider + key + live model auto-detect (`/api/models`) → strategy + basket chips →
  Pimlico key (paper-mode default) → Telegram token with inline test → "what's next" (grant / faucet /
  doctor). Every step is optional; blanks keep the paper-mode defaults.
- Finish or skip writes `settings.webOnboarded`, so neither surface ever nags a configured install.
- `Field` + `secretPlaceholder` extracted to `web/src/components/setup-fields.tsx` so `/settings` and
  the wizard use identical field chrome. Mounted in `app/page.tsx`; `.onboard-*` styles in `globals.css`.

## Terminal channel choice

- `cli/bin.mjs start` — on a fresh install it asks, in the same style as the installer's local/Docker
  pick:
1. in the browser — the dashboard wizard (localhost:3100)
2. right here in the terminal (merrymen onboard)
Choice 2 runs the onboard flow inline; no TTY (headless/Docker) skips to the browser path.
- `onboard()` writes `webOnboarded: true`, so a terminal setup never doubles up with the wizard.

## Shared marker + safe writes

- `settings.webOnboarded` (new) — the one flag both surfaces read, so setting up on either silences both.
- Both writers (`/api/settings` PUT and the CLI) now write `settings.json` atomically (temp file +
rename, 0600) — a crash mid-write can't leave a truncated config the worker silently reads as defaults.

Everything entered on either surface lands in the same `~/.merrymen/settings.json`, so each shows up on
the other.

## Screenshots

**1. First-run wizard on the dashboard**
1
Screenshot_20260806_152713
2
Screenshot_20260806_152816
3
Screenshot_20260806_152833
4
Screenshot_20260806_152850
5
Screenshot_20260806_152959
6
Screenshot_20260806_153011

**2. First-run channel prompt in the terminal**
Screenshot_20260806_153027

## Tests

`npm run typecheck` (worker + web) clean · `npm run build` (web) clean · `npm test` — 692 pass, 0 fail.


millw14:
First, credit where it's due. I came into this expecting the classic atomic-write bug — temp file created with default permissions, rename carries the temp's mode, and `settings.json` silently goes from 0600 to world-readable while looking like a robustness improvement. **You handled it.** Both paths pass an explicit `{ mode: 0o600 }` *and* chmod: the CLI chmods the temp before the rename, the web route chmods the target after it. The temp is `${target}.tmp`, same directory, so no cross-device rename either. I was wrong to expect that one, and it's the detail most people miss.

That said, the atomic write has a real regression, and the wizard has three.

### 1 · The rename is *less* reliable than the write it replaced, on Windows

I reproduced this on Node v22.17.0 / win32, holding one plain read handle open on the target:

```
OLD  direct writeFileSync          OK             stray .tmp: false
NEW  writeFile(tmp)+rename         FAILED EPERM   stray .tmp: true
```

Windows refuses a rename over a file any process holds open. On Linux that's fine; on Windows it's the common case — an editor, a backup agent, antivirus, or merrymen's own reader. Neither call site (`cli/bin.mjs:122-129`, `web/src/app/api/settings/route.ts:458-461`) has a `try`/`catch`, so the save is simply lost: in the CLI it's a stack trace and exit 1 with the secrets the user just typed gone, and in both cases a complete `settings.json.tmp` is left behind — 0600, so not an exposure, but a second copy of every API key and the bot token sitting on disk unnoticed.

Worth being precise about the tradeoff: on POSIX this change is a genuine improvement, and on Windows it converts a rare torn write into a common lost write. Retry the rename a few times with a short backoff, `catch` and clean up the temp on failure, and surface the error rather than throwing out of the CLI.

### 2 · Only two of the three settings writers were converted

The one most likely to run unattended still does a bare truncating write, which is exactly the case atomicity is for. Either convert it too or say in the PR why it's excluded — as it stands the invariant the PR establishes isn't actually established.

### 3 · The wizard probes the model API on **every** dashboard load, with your stored key

`OnboardWizard` decides not to render at `OnboardWizard.tsx:186-187` — but that's in the render body, *after* the hooks. A component that returns `null` still mounts and still runs its effects, and `page.tsx:75` mounts it unconditionally on `/`, which is also the PWA start page. So on a fully-onboarded install where the wizard shows nothing, every page load still fires the debounced `/api/models` effect and the local server sends the stored Groq key to `api.groq.com`.

It goes to the legitimate provider, so this isn't a leak — but sending a credential off the machine should be something the user asked for, not something that happens because they opened a page. Hoist the show/hide decision above the hooks and make it the first line of the effect.

### 4 · `merrymen start` can hang forever

The new first-run prompt blocks on interactive input with no non-interactive escape. In any TTY-attached but unattended context — the Windows auto-start shim at boot, a terminal multiplexer, a service wrapper — `merrymen start` waits for a keypress that never comes, and neither the dashboard nor the worker launches. Detect non-interactive (`!process.stdin.isTTY`, `CI`, `MERRYMEN_NONINTERACTIVE`) and fall through to the default, or gate the prompt behind `onboard` where a human is definitionally present.

### 5 · The first-run gate is blind to env configuration

It looks only at `~/.merrymen/settings.json`, but env vars are a first-class config source — and file values win over env. So a Docker or systemd install configured entirely through the environment is a live, working agent that gets shown the first-run wizard. Check resolved config rather than the file's existence.

---

Smaller, worth a pass: both writers share one constant temp filename, so two concurrent saves can publish a truncated file; the final "what's next" panel unmounts itself 600 ms after appearing, so nobody reads the grant/faucet/doctor steps; pasting a bot token silently flips `telegramEnabled` to true and starts the outbound long-poll with no visible toggle; and there's an HTML entity (`&apos;`) inside a plain JS string literal that will render literally.

The direction is right and the wizard is a real gap in the product. It's items 1, 3 and 4 that would bite people.

millw14 (commented):
Thanks for this. Two things I want independent of the feature:

- Extracting `Field`/`secretPlaceholder` into `setup-fields.tsx` so `/settings` and the wizard can't drift is exactly right.
- Making both settings writers **atomic** (temp file + rename) is a real robustness win on its own — a crash mid-write leaving a truncated `settings.json` that the worker reads as defaults is a bug waiting to happen, and I'd want that change even if the wizard never merged.

I checked the thing most likely to make this dead on arrival — whether `values` from `GET /api/settings` is merged with defaults, which would make `configured()` always true and the wizard never render. It isn't; `readStored()` returns raw disk contents. Good.

One blocker:

**`bundlerUrl` is missing from both gates.** `hasMeaningfulConfig()` and `configured()` check `bundlerApiKey`, but `settings.ts:22` documents `bundlerUrl` as taking *precedence* over it — it's the other supported way to configure the bundler. So a user who pasted a full Pimlico URL, or who set up through `/grant` alone, upgrades and gets a full-screen onboarding overlay across their working dashboard plus a blocking prompt from `merrymen start`. Please add `bundlerUrl` to both, and consider "a grant exists" as a third signal — someone with a funded agent has unambiguously onboarded.

Three more:

1. **Rendering bug:** in the done state, the string `"...but something didn&apos;t save:"` puts an HTML entity inside a *JavaScript string*, not JSX text — it renders literally as `didn&apos;t`. Use a real apostrophe.

2. **The first-run terminal prompt can spin forever.** `while (choice !== "1" && choice !== "2")` has no EOF handling: if stdin reports a TTY and then closes, `ask()` resolves `""` in a tight loop. It needs a bail-out, and probably a non-interactive escape too — `merrymen start` is what a supervisor and the desktop app run, and it must never be able to block on a prompt.

3. **No tests.** New settings field, new API surface, new CLI branch. At minimum: `webOnboarded` round-trips through `PUT`/`GET`, and the gate treats a `bundlerUrl`-only config as onboarded.

Also two of the new files are missing trailing newlines.

millw14:
To land this: the `bundlerUrl` gate fix is the one I'm holding on — as it stands, anyone who configured their bundler by URL rather than by API key gets a full-screen wizard over a dashboard they've already set up. The `didn&apos;t` render bug, the EOF spin in the terminal prompt, and a couple of tests would all be good to have in the same push.

This branch is also about three weeks behind `main` — worth rebasing before you push, so CI runs against current code:

```bash
git fetch upstream
git rebase upstream/main
git push --force-with-lease
```

Pushing to the same branch updates this PR in place.

Independent of the wizard: the atomic settings write (temp file + rename) is a good change on its own merits. If you'd rather split it into its own small PR, I'll take that immediately and we can let the wizard take its time.

aor-rex:
Fixes all review points, rebased onto `main` (`b525450`).

### Blockers

**`bundlerUrl` gate** — `hasMeaningfulConfig()` (`cli/bin.mjs:163`) and `configured()` (`OnboardWizard.tsx:29`) now check `bundlerUrl` (takes precedence per `settings.ts:22`) and `grant.json` existence, so a URL-configured or funded agent doesn't get the full-screen wizard.

**Render bug** — `OnboardWizard.tsx:378` `didn&apos;t` → `didn't` (was a JS string inside `{}`, not JSX — rendered literally).

**Prompt spin** — `merrymen start` (`cli/bin.mjs:232,656`) — `makePrompter` resolves `null` on `rl.close`, loop bails on `null`/`""` to browser default, and gate skips when `!isTTY || CI || MERRYMEN_NONINTERACTIVE`.

### Windows-safe atomic writes

All three writers now use unique `tmp.${pid}.${rand}`, retry `rename` 5× on `EPERM`/`EBUSY`/`EACCES` with backoff, `rmSync(tmp)` on failure, and surface a clean error:

- `cli/bin.mjs:122` — exits with `couldn't save settings` (no stack, temp cleaned)
- `web/src/app/api/settings/route.ts:472` — returns `500 {errors: [...]}` and removes temp
- `worker/src/settings.ts:313` `patchSettingsFile` — was bare truncating `writeFileSync` (the unattended Telegram path) — now atomic

Constant `${target}.tmp` race fixed.

### Wizard probe + env gate

- **Hoisted gate** — `OnboardWizard.tsx:108` first lines `if (view.webOnboarded || configured(view)) return` before the debounced `POST /api/models` — hidden wizard never sends the stored key. `page.tsx:75` still mounts unconditionally but effects early-return.
- **Env-aware** — `cli/bin.mjs:170` and `route.ts:123` check `MERRYMEN_BUNDLER_URL`, `MERRYMEN_BUNDLER_API_KEY`, `GROQ_API_KEY`, `ANTHROPIC_API_KEY`, `MERRYMEN_LLM_API_KEY`, `MERRYMEN_TELEGRAM_BOT_TOKEN`, `MERRYMEN_LLM_PROVIDER`, `MERRYMEN_STRATEGY` + `grant.json`; `SettingsView.envHasConfig` exposed to wizard.

### Small fixes

- Remove `600ms` auto-hide (`OnboardWizard.tsx:187`) — "what's next" stays until navigation.
- `telegramEnabled` no longer silent — `OnboardWizard.tsx:69,336` checkbox `start the bot now`, `cli/bin.mjs:587` `[Y/n]` prompt.
- Trailing `\n` on `OnboardWizard.tsx` + `setup-fields.tsx`.

**Tests:** `worker/src/settings.test.ts:215` — `bundlerUrl`-only counts as onboarded, `webOnboarded` round-trips `PUT`→`GET` (`true`/`false`/`null` clears), env-configured install counts. `935/935` green, `typecheck` clean, `node --check cli/bin.mjs` ok.

millw14:
**This needs rebuilding rather than rebasing — the files it edits no longer exist on `main`.**

The terminal rewrite (#96 `feat/fomo-shell`, plus #103) deleted the surfaces this PR is built on, including `web/src/app/page.tsx`, `web/src/app/globals.css`, `web/src/app/settings/page.tsx` and `web/src/components/OnboardWizard.tsx`.

GitHub reports CONFLICTING, but conflict resolution will not recover this — the dashboard wizard it extends has been replaced.

**What to do:** compare against the onboarding that shipped in #96 and decide whether this is already covered. If it is, close it. If something is genuinely missing, re-cut it against the current terminal. Worth settling alongside #11, which is the Telegram half of the same idea.

aor-rex:
Rebased onto current `main` — 5 commits replayed, pushed over this same branch. Three conflicts, all both-sides: `settings/route.ts` kept main's tenant-store branching + ignored-keys reporting with the atomic write underneath (upgraded to the Windows-safe variant: unique tmp name, rename retry, cleanup + 500 on failure); the old dashboard `page.tsx`/`settings/page.tsx` are gone on main (terminal App architecture now), so the wizard mounts in `(app)/layout.tsx` and its styles moved from quarantined `legacy.css` to loaded `terminal/forms.css` — without that move it would render unstyled. Local: `npm ci` + typecheck (worker/web/browser) + full suite (4880 pass / 0 fail). CI on the rebased tip: `app` and `contracts` green.

millw14 (commented):

**[P2] EOF now crashes terminal onboarding instead of finishing safely** — `cli/bin.mjs:236`. `makePrompter().ask()` now resolves `null` on close, but existing `onboard()` callers continue invoking `.trim()` directly, including the new Telegram enable prompt at line 589. EOF during those questions produces a TypeError and abandons the unsaved setup. Handle a cancelled/closed prompt consistently at every caller, or return a result that callers already safely consume. Test EOF during onboarding, not only the new first-run channel picker.

Validation: 23 settings tests passed. The added onboarding gate tests duplicate the helper logic rather than exercising the CLI or rendered wizard. No complete browser onboarding session was run.

aor-rex:
Confirmed and fixed — thank you. EOF crashed (and in one path hung) exactly as described.\n\n**Fix:** closed prompts now resolve `""` — the documented blank — instead of `null` (ask) or hanging forever (askSecret, which had no close handler at all and also leaked its output override). I audited every caller rather than trusting the shape: keys/secrets/addresses are all `if (value)`- or regex-guarded, picks treat blank as keep-current, kill requires `y`, sweep requires the literal word — so EOF inherits precisely the semantics each prompt text promises. The channel-picker's existing null-check still takes the same branch (empty string hits its `=== \"\"` eof path identically).\n\n**Testability fix included:** `makePrompter` extracted to `cli/prompter.mjs` (bin.mjs executes on import and can never load under a test runner) with new `cli/prompter.test.mjs` — immediate-EOF resolves blank, secret doesn't hang (2s hang-guard assertion), normal answers flow, and the exact `.trim()` caller shape is exercised. Test glob extended to `cli/*.test.mjs`.\n\n**Evidence:** new CLI tests green; full suite green locally; `app`+`contracts`+`gateway` green on the pushed tip.

millw14 (commented):

**Needs changes.** Returning an empty string fixes the first pending question's `.trim()` failure, but EOF still breaks the multi-question onboarding flow.

**[P2] Remember that readline has closed before asking the next question** — `cli/prompter.mjs:17–30`. After EOF, the active question resolves blank and onboarding proceeds to its next `p.ask`/`p.askSecret`. That call attaches a new close listener to an already-closed interface and calls `rl.question`, which throws `ERR_USE_AFTER_CLOSE`; the close event will not fire again. Keep a closed flag and short-circuit subsequent calls to the documented cancellation result, or stop onboarding explicitly. Cover EOF followed by another ask and another askSecret; all new tests stop immediately after the first EOF response.

Verification: source/test inspection plus an independent Node standard-library check confirming `question()` after `close()` raises `ERR_USE_AFTER_CLOSE`. No contributor code was executed. Existing CI is evidence for the old base only; resolve current main conflicts and rerun checks. No automatic merge authorization.

PR #13, opened 2026-08-07
fix: make PC-control commands actually work on Linux — desktop-aware toolchains + install/daemon offers

aor-rex:
## Problem

On a Linux desktop the PC-control commands didn't work at all. The platform
layer hardcoded X11 tools everywhere: typing/hotkeys ran `xdotool`, screenshots
ran `scrot`/`import`/`gnome-screenshot`, clipboard ran `xclip`. It never
detected the session type, so on a Wayland desktop (e.g. KDE Plasma) every one
of those binaries either won't run or silently does nothing — /type, /hotkey,
/screenshot and the clipboard commands all just failed with a generic error.

Two more gaps with no recovery path:
- a genuinely-missing tool produced a cryptic failure with no offer to install
  it (and no package-manager awareness);
- if `ydotool` was installed but its daemon (`ydotoold`) wasn't running, typing
  failed with no way to fix it.

## Change

- `worker/src/pc/platform.ts` — session-aware Linux support:
  `sessionType()` distinguishes wayland / x11 / headless, and toolchains pick
  the tool that actually works in the current environment (KDE/GNOME Wayland →
  `ydotool`, wlroots → `wtype`, X11 → `xdotool`; `grim`/`import` for capture;
  `wl-paste`/`xclip` for clipboard). Headless boxes fail fast with a clear
  reason instead of a toolchain hunt.
- Install + daemon-start offers, parked behind /confirm (or ✅ tap):
  `worker/src/telegram/{pc,service,executor}.ts`. A /confirm that resolves one
  ask and parks another re-attaches the ✅/✖ buttons to the fresh offer;
  `install_tool` agent tool added so the agent can offer installs mid-task
  (park-only; runs only after owner confirms).
- Pure decision logic (`typingOutcome` / `hotkeyOutcome`) so a failing run picks
  the right next step — offer the missing tool, offer the daemon start, or
  report plainly — and never re-offers a tool that's already installed (the
  /type infinite-loop regression).

## Requirements

- Installs run under `sudo -n <pm>` with `--noconfirm` (pacman) — so the
  📦 offer only appears when the owner has passwordless sudo for that package
  manager (probed with `sudo -n <pm> -V`, matching the binary a NOPASSWD rule
  whitelists, not a bare `sudo -n true`); brew needs no sudo. Without
  passwordless sudo no install is offered and the failure is reported plainly.
- Installs are gated behind their own `install` PC capability (default off,
  disclosed in /pc and the dashboard): the offer, the ✅ tap and /confirm all
  refuse when it is off, and /confirm re-vets through `pcRefusal` so revoking
  it after parking refuses too. Never offered for already-installed tools.
- One-time setup: `/pc` names the missing packages for manual install and, as
  an alternative, describes a NOPASSWD rule for exactly the user merrymen runs
  as (never `%wheel`, never a copy-pasteable command).
- Daemon starts need no sudo: `systemctl --user enable --now ydotool`.

## Tests

`npm test` — 4938 pass, 0 fail locally. `npm run typecheck` clean
(worker + web + browser). Required CI checks (`app`, `contracts`) re-running
on the rebuilt branch.

## System packages this can install (on /confirm, as root)

`wl-clipboard`, `xclip`, `grim`, `gnome-screenshot`, `spectacle`, `scrot`,
`maim`, `imagemagick`, `wtype`, `ydotool`, `xdotool`, `playerctl`, `libnotify`,
`wireplumber`, `alsa-utils`. Runtime also invokes `sudo`, the distro package
manager, `systemctl --user`, `pactl`, `wpctl` and `pgrep`. No npm dependencies.

> Rebuilt Sept 2026: Linux-only work cherry-picked from `698a916` onto current
> `main` (the old tip carried #11's files); pushed over this same branch.

### DIFF
--- a/web/src/app/globals.css
+++ b/web/src/app/globals.css
@@ -569,6 +569,54 @@ button { font: inherit; cursor: pointer; }
 }
 .tg-cta-btn:hover { background: #1c93d0; }
 
+/* ---------- telegram link modal (global first-link onboarding) ---------- */
+.tg-modal-backdrop {
+  position: fixed; inset: 0; z-index: 100;
+  display: flex; align-items: center; justify-content: center;
+  padding: 20px;
+  background: color-mix(in srgb, var(--bg) 62%, transparent);
+  backdrop-filter: blur(3px);
+  animation: tg-fade 160ms ease-out;
+}
+.tg-modal {
+  position: relative;
+  width: min(380px, 100%);
+  display: flex; flex-direction: column; gap: 12px;
+  padding: 22px 20px 20px;
+  background: var(--bg-2);
+  border: 1px solid var(--border-2);
+  border-radius: var(--radius);
+  box-shadow: 0 18px 50px rgba(0, 0, 0, 0.5);
+  animation: tg-pop 180ms ease-out;
+}
+.tg-modal-close {
+  position: absolute; top: 10px; right: 12px;
+  width: 28px; height: 28px; line-height: 1;
+  border: none; background: none; color: var(--text-faint);
+  font-size: 22px; cursor: pointer; border-radius: 6px;
+}
+.tg-modal-close:hover { color: var(--text); background: var(--bg-3); }
+.tg-code-row { display: flex; align-items: stretch; gap: 8px; }
+.tg-code {
+  flex: 1;
+  font-family: var(--mono); font-size: 17px; font-weight: 700;
+  letter-spacing: 4px; text-align: center;
+  color: var(--green-soft);
+  background: var(--bg-3); border: 1px solid var(--border-2);
+  border-radius: 8px; padding: 10px 8px;
+  user-select: all;
+}
+.tg-copy-btn {
+  font-family: var(--mono); font-size: 12px; font-weight: 700;
+  color: var(--text); background: var(--bg-3);
+  border: 1px solid var(--border-2); border-radius: 8px;
+  padding: 0 12px; cursor: pointer; white-space: nowrap;
+}
+.tg-copy-btn:hover { border-color: var(--green-deep); color: var(--green-soft); }
+.tg-modal-note { font-size: 10.5px; color: var(--text-faint); }
+@keyframes tg-fade { from { opacity: 0; } to { opacity: 1; } }
+@keyframes tg-pop { from { opacity: 0; transform: translateY(8px) scale(0.98); } to { opacity: 1; transform: none; } }
+
 /* ---------- the wall (trust layer) panel ---------- */
 .wall-panel { display: flex; flex-direction: column; gap: 10px; }
 .wall-sub { font-size: 11.5px; color: var(--text-dim); line-height: 1.55; }
--- a/web/src/app/layout.tsx
+++ b/web/src/app/layout.tsx
@@ -2,6 +2,7 @@ import type { Metadata, Viewport } from "next";
 import { Hanken_Grotesk, JetBrains_Mono } from "next/font/google";
 import "./globals.css";
 import { RegisterSW } from "@/components/RegisterSW";
+import { TelegramLinkModal } from "@/components/TelegramLinkModal";
 
 // The merrymen.dev typefaces — used on the setup/settings screens (.setup-look)
 // so onboarding feels like the website; the trading terminal keeps its own fonts.
@@ -50,6 +51,7 @@ export default function RootLayout({ children }: { children: React.ReactNode }) 
       <body>
         {children}
         <RegisterSW />
+        <TelegramLinkModal />
       </body>
     </html>
   );
--- a/web/src/components/TelegramLinkModal.tsx
+++ b/web/src/components/TelegramLinkModal.tsx
@@ -1,0 +1,134 @@
+"use client";
+
+import Link from "next/link";
+import { useEffect, useState } from "react";
+import type { TelegramStatus } from "@/app/api/telegram/route";
+
+/**
+ * Global first-link onboarding popup. Appears on any dashboard/settings page as
+ * soon as the user has a valid Telegram token AND the bot is reachable, but has
+ * not linked an owner chat yet. It guides them to the one-time /link code and
+ * auto-closes the moment the worker confirms a link (ownerId appears).
+ *
+ * Deliberately NOT persisted: dismissal lasts only for the lifetime of the open
+ * page. A fresh visit re-shows the popup until the bot is actually linked, so
+ * the "I can't see the code" gap can't be dismissed away forever.
+ */
+export function TelegramLinkModal() {
+  const [tg, setTg] = useState<TelegramStatus | null>(null);
+  const [dismissed, setDismissed] = useState(false);
+  const [copied, setCopied] = useState(false);
+
+  useEffect(() => {
+    let alive = true;
+    const load = async () => {
+      try {
+        const r = await fetch("/api/telegram");
+        if (alive && r.ok) setTg((await r.json()) as TelegramStatus);
+      } catch {
+        /* keep last */
+      }
+    };
+    load();
+    const id = setInterval(load, 5000);
+    return () => {
+      alive = false;
+      clearInterval(id);
+    };
+  }, []);
+
+  const open = !!tg && tg.hasToken && tg.connected && tg.ownerId === null && !dismissed;
+
+  useEffect(() => {
+    if (!open) return;
+    const onKey = (e: KeyboardEvent) => {
+      if (e.key === "Escape") setDismissed(true);
+    };
+    window.addEventListener("keydown", onKey);
+    return () => window.removeEventListener("keydown", onKey);
+  }, [open]);
+
+  if (!open || !tg) return null;
+
+  const botLink = tg.botUsername ? `https://t.me/${tg.botUsername}` : null;
+  const code = tg.linkCode;
+  const preparing = !!tg.enabled && !code;
+  const needsTurnOn = !tg.enabled && !code;
+
+  const copyCode = async () => {
+    if (!code) return;
+    try {
+      await navigator.clipboard.writeText(code);
+    } catch {
+      const ta = document.createElement("textarea");
+      ta.value = code;
+      ta.style.position = "fixed";
+      ta.style.opacity = "0";
+      document.body.appendChild(ta);
+      ta.select();
+      document.execCommand("copy");
+      document.body.removeChild(ta);
+    }
+    setCopied(true);
+    setTimeout(() => setCopied(false), 1500);
+  };
+
+  return (
+    <div
+      className="tg-modal-backdrop"
+      onClick={(e) => {
+        if (e.target === e.currentTarget) setDismissed(true);
+      }}
+    >
+      <div className="tg-modal" role="dialog" aria-modal="true" aria-label="Link your Telegram bot">
+        <button type="button" className="tg-modal-close" aria-label="Dismiss" onClick={() => setDismissed(true)}>
+          ×
+        </button>
+        <div className="tg-card-title">💬 Link your Telegram</div>
+
+        {code ? (
+          <>
+            <p className="tg-card-sub">
+              Open <b>{tg.botUsername ? `@${tg.botUsername}` : "your bot"}</b> and send{" "}
+              <code>/link {code}</code> — the first message that claims it makes you its owner.
+            </p>
+            <div className="tg-code-row">
+              <code className="tg-code">{code}</code>
+              <button type="button" className="tg-copy-btn" onClick={copyCode}>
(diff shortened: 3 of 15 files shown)


millw14:
This is a lot of real work, and the parts that are good are genuinely good: `resolveToolChain` picking a working binary from an ordered list, `PACKAGE_MAP` with a stated rule that an install you aren't sure about must never be offered, `sudo -n` probing rather than assuming, and tests for the resolution logic. Linux PC control being broken is a real bug and this clearly comes from actually using it.

**First, scope.** `git merge-base --is-ancestor pr11 pr13` returns true — this branch **contains #11**. So everything in my review over there applies here too, most importantly that the ✅ Confirm button carries no identity for the action it was posted with and executes whatever is parked at tap time. That matters more in this PR than in #11, because #13 adds new *spontaneous* writers to the pending slot. Please rebase onto main once #11 lands rather than carrying it, so this diff is only the Linux work.

Two things to fix here beyond that.

### 1 · Blocker — the bot tells the owner to give the wheel group passwordless root

`/pc` prints this when tools are missing:

```
some tools are missing — set up scoped NOPASSWD (e.g. `echo '%wheel ALL=(ALL) NOPASSWD: /usr/bin/pacman' | sudo tee /etc/sudoers.d/merrymen-pm`) and I can install them on /confirm.
```

It's called "scoped", and it is scoped to a binary — but the binary is a package manager, so the scope is illusory. `pacman -U ./anything.pkg.tar.zst` installs a local package whose install scripts run as root. `apt-get` will happily install a local `.deb`, and `-o APT::Update::Pre-Invoke::` runs arbitrary commands. dnf and apk are no different. A passwordless package manager **is** passwordless root, one command later.

And it's granted to `%wheel` — every account in that group, not the one merrymen runs as.

The thing I'd push back on hardest isn't the rule itself; it's that **the software is talking its owner into weakening their machine**, in a chat message, to enable auto-installing screenshot tools. A trading agent that holds your wallet key should be the last program on the box asking for that. Someone who wants passwordless pacman can decide that on their own; it shouldn't arrive as a copy-pasteable one-liner from a bot.

**Suggested instead:** print the *package names* and let the owner install them however they like — `sudo pacman -S wl-clipboard grim`. That's one manual step, once, and it costs the feature almost nothing. If auto-install stays, at minimum scope the rule to the merrymen user rather than `%wheel`, and drop the word "scoped" from the copy since it isn't.

### 2 · High — the "never re-offer an installed tool" fix only covers the typing path

The PR describes not re-offering tools that are already present, but `installPlanFor()` has no already-installed guard, and `captureLinux` / `setVolumeLinux` call it without one — so the screenshot and volume paths keep offering an install for something already on the box. Worth applying the same check there.

---

Also worth noting for the body rather than as a defect: this installs real system packages as root on confirm — `wl-clipboard`, `xclip`, `grim`, `gnome-screenshot`, `spectacle`, `scrot`, `maim`, `imagemagick`, `wtype`, `ydotool`, `xdotool`, `playerctl`, `libnotify`, `wireplumber`, `alsa-utils` — and invokes `sudo`, the distro package manager, `systemctl --user`, `pactl`, `wpctl` and `pgrep` at runtime. No npm dependencies, which is the right call. But the list of *system* packages a merged PR would cause to be installed belongs in the description where a reviewer sees it.

**Honest limit on this review:** I have no Linux box here, so I read the code paths rather than running them. The toolchain resolution and the install-plan construction I can reason about from the diff; whether the resulting commands actually work on Wayland versus X11 on a real distro, I can't confirm and haven't claimed to.

At 1938 lines this is also very large for a fix. If the sudoers piece comes out and #11 is rebased away, what's left is a coherent "Linux toolchain detection" change that I'd be glad to take.

millw14:
Heads-up: this branch now conflicts with `main`, and the conflict is my doing, not yours.

[`eaeb75a`](https://github.com/millw14/merrymen/commit/eaeb75a) added a `/depth` read command, which touched the two spots in `worker/src/telegram/service.ts` your branch also edits:

- `TelegramServiceDeps` gained a `readDepth: (symbol: string) => Promise<string>;` member,
- `cmdDeps.reads` gained a `depth: (symbol: string) => deps.readDepth(symbol),` line.

The resolution is additive — keep both sides. Your changes and mine don't overlap semantically; they just landed on adjacent lines of the same interface and the same `reads` block. (`executor.ts`, `interpreter.ts` and `interpreter.test.ts` auto-merge cleanly; `service.ts` is the only file needing a hand.)

A plain `git merge origin/main` (or rebase, your preference) with those two hunks kept both-sides should do it. CI on `main` is green with the depth command in, so once the conflict clears the required checks should pass as they did on your last run. Sorry for the churn — the review itself stands as posted.

millw14 (commented):
Linux PC control genuinely doesn't work today, so I want this — and the failure modes you've handled say you actually hit them: preferring the desktop's native tool over looping on an installed-but-unusable `wtype`, `--noconfirm` because `sudo -n` silences sudo's prompt but not pacman's, `sudo -n` so it can never sit blocked on a password. The argv-from-a-fixed-map discipline is the right call and it's held consistently.

But this moves a privilege boundary — merrymen can now run `sudo -n apt-get install -y <pkg>` — so I'm going to be fussy about two things.

**1. The re-vet at confirm time is weaker for installs than for everything else.** In the confirm branch, `install` and `service` get their own arm that re-checks only `deps.pcControlEnabled`, while every other PC kind goes through `pcRefusal({ kind: p.kind }, deps)`, which also checks the per-capability set. So if a user parks a `/type`-triggered ydotool install and revokes the `keyboard` capability before confirming, the install still runs. That's the exact class of bug the 0.10.2 work fixed for the other kinds — and it's backwards here, because `apt-get install` runs maintainer scripts **as root**, which is strictly more dangerous than the capability that triggered the offer.

**2. Installing isn't a capability, so nothing surfaces it.** There's no `install` entry in `PC_CAPABILITIES`, so `/pc` never mentions it and the dashboard's remote-execution warning (which covers shell and keyboard) doesn't either. A user enabling PC control cannot see that they're also granting root package installs. I'd like it as its own capability, defaulting off — that also gives (1) something concrete to re-vet against.

Smaller:

- `pcResultReply` reads like the shared helper for this, but only `typeText` uses it — every other call site inlines the `(r.missing ? offer(...) : null) ?? fail(...)` form. Pick one.
- Worth keeping and worth saying: moving `note("warn", "typed into the active window")` after the failure check means the log stops claiming a keystroke that never landed. Good catch.

**Process ask:** this isn't a clean stack on #11. It re-includes `TelegramLinkModal.tsx` and the `api.ts` changes, and carries its own divergent copies of `describePending`/`PendingAction` — but without #11's nonce field or the pending-blocks-park change. Merging both as they stand will conflict badly in `executor.ts` and `service.ts`. Once #10 and #11 are in, please rebase onto them and drop the duplicated files, so this PR is just the Linux toolchain work.

millw14:
Sequencing on this one, since it's the most tangled:

1. I'm landing #10 first, then #11.
2. Once those are in, rebase this onto `main` and drop the files it shares with #11 — `TelegramLinkModal.tsx`, the `api.ts` changes, and the divergent copies of `describePending`/`PendingAction`. That should leave this PR as just the Linux toolchain work, which is what I want to review it as.
3. The two asks from my review: re-vet installs against the per-capability set rather than only the master switch, and give `install` its own entry in `PC_CAPABILITIES` so `/pc` and the dashboard actually disclose it.

```bash
git fetch upstream
git rebase upstream/main
git push --force-with-lease
```

Pushing to the same branch updates this PR in place — no need to open a new one.

No rush on the timing; the rebase will be much less painful after #11 lands than before.

millw14:
**Still wanted, but it needs a rebase before it can be judged.**

CONFLICTING against current `main`, and the last CI run here is from August. The PC-control surface has changed since, so please rebase and let CI re-run.

**What to do:** rebase onto current `main`, confirm the desktop-detection logic still matches how `cli/bin.mjs` dispatches today, and make sure `app` and `contracts` are both green. Please also confirm this does not overlap #106, which is the other Linux-facing PR open right now.

---

aor-rex:
Rebuilt the branch per the sequencing ask: the old tip (`698a916`, August `main`, carrying #11's files) is replaced by the Linux-only work cherry-picked onto current `main`, pushed over this same branch — no new PR. `TelegramLinkModal.tsx` and the divergent `describePending`/`PendingAction` copies are gone; the `api.ts` button/markup support stays because `main` doesn't have it (needed for the ✅/✖ confirm path — flag if you'd rather it land separately).

Review responses, all in the single rebuilt commit:
1. Installs/service-starts re-vet through `pcRefusal` at confirm — revoking the new `install` capability after parking refuses (covers the /type-triggered-ydotool case).
2. `install` is its own entry in `PC_CAPABILITIES`, default off, disclosed in `/pc` and the dashboard warning (which now names root package installs as the most privileged thing on the page).
3. Sudoers hint no longer names `%wheel`, no longer says "scoped", no longer a copy-pasteable `tee` command — it names the merrymen user and offers manual install as the alternative. Full removal of auto-install is a yes/no away if you prefer.
4. `installPlanFor` refuses already-installed tools centrally, so typing/screenshot/volume inherit it.
5. Smaller items: all `pc.ts` failure paths share `pcResultReply`; the typed-into-window log line already sits after the failure check.

One thing the suite caught during the rebuild: the resolution had initially dropped `main`'s `linkedChats` persistence hunk — restored (peer-keyed for the shared linkDep) with a source-content test; full audit of the other deletions done, all intended.

`npm run typecheck` clean (worker + web + browser), `npm test` 4938 pass / 0 fail locally. The system-package list is now in the PR description as asked.

millw14 (commented):

1. **[P1] Inline confirmation buttons have no inbound path** — `worker/src/telegram/service.ts:1098` defines `handleCallback`, but the poll loop never calls it. `worker/src/telegram/api.ts` still requests only `allowed_updates: ["message"]` and returns only parsed messages. The newly displayed Confirm/Cancel buttons therefore cannot reach the executor. Request and parse callback queries, dispatch them through the same authorization checks, and test a complete update → callback → pending-action flow.
2. **[P2] Every reply is sent twice, with the second bypassing reasoning filtering** — `worker/src/telegram/service.ts:1086`. The existing block already sends `strippedReply` and records history. The new block records history again and sends raw `reply` with the keyboard. This duplicates ordinary replies and can publish text the first block intentionally stripped. Consolidate to one filtered send and one history write, attaching the keyboard there.

Validation: 129 focused platform/interpreter/PC tests passed. These do not exercise the complete service poll-and-send path. Linux desktop tooling itself was not exercised on this Windows host.

aor-rex:
Both findings confirmed against the code and fixed — thank you, these were real. P1 was the worse miss: the buttons rendered while taps died silently, green CI notwithstanding.\n\n**P1 — taps now reach the executor.** `getUpdates` subscribes to `callback_query` (was message-only), parses taps into `TgCallback` (offset advances over them; context-free taps drop), and the poll loop dispatches through `handleCallback` under the same allowlist rule as messages — extracted as pure `isCallbackSenderAllowed` precisely so this gate is unit-tested (an untested gate is how the buttons shipped wired to nothing). No /link exception for taps: a tap can only resolve, never authorize.\n\n**P2 — one send.** Consolidated to a single filtered send + single history write with the keyboard attached; the raw second send is gone.\n\n**Evidence:** new mixed-batch parsing test (tap/message/junk split, offset, subscription assertion) + tap-authz matrix; targeted telegram suites green; full suite 5189 pass / 0 fail locally; `app`+`contracts`+`gateway` green on the pushed tip. (One note: `recovery-keyless-sweep` flakes under parallel-suite load in this environment — passes alone 5/5 repeatedly, zero shared imports with the touched files; unrelated.)

millw14 (commented):

**Needs changes.** Callback queries are now requested and dispatched, and the duplicate raw Telegram send is removed. Connecting that previously dead path exposes a confirmation-binding issue.

**[P1] Bind each button to the action shown in its message** — `worker/src/telegram/service.ts:1102–1119`. Buttons carry only `confirm`/`cancel`, and the handler checks only whether the sender's shared pending slot exists. Park transfer A, then park transfer B (the executor explicitly uses latest-ask-wins), and tap A's old Confirm button: it confirms B. The same mismatch can cross action kinds. Include an action nonce or message binding and reject superseded buttons before calling the executor; test A → B → tap A, including a transfer or shell action.

**[P2] History is still written twice** — immediately before `handleCallback`, the original awaited `pushHistory(...stripThinkingBlock(...))` remains alongside the new `pushHistory(...safeReply...)`. Keep one filtered history write as well as one send, so later model context is not duplicated.

Verification: source inspection and existing GitHub evidence only; no contributor code or live Telegram commands executed. The PR conflicts with current main and remains outside the automatic merge lane.

PR #17, opened 2026-08-31
Re-sign and change chain from the grant page, without the side door

millw14:
Independent of #16 — branches off `main`, touches only `web/src/app/grant/page.tsx`.

## The gap

The only renewal control lives inside the expiry notice, which returns `null` unless the key has under three days left (`page.tsx:1073`). So an owner wanting to re-sign a **healthy** key has no button — while the panel directly above tells them to do exactly that, twice: *"Renew below to widen it"* and *"Renewing the key re-signs it without the pair."* The copy points at a control that usually isn't rendered.

The routes that did work were all side doors: wait for near-expiry, add a token the key doesn't cover, or go through a flow labelled **"switch to another wallet"** and paste the owner key back in.

**This isn't cosmetic.** The wall changes. A ZeroDev policy contract turned out to have no code on this chain (#14), so every key signed before that fix carries a pointer into empty space — and fixing the wall does nothing for a signature already made. Re-signing is the only remedy, and it was the one thing the page would not let you do.

## The change

A **"re-sign this key"** section in the loaded-grant panel, immediately after "what this key carries" — where the copy already pointed. Same `renewKey()` call, no expiry gate.

**Caps are editable there, deliberately.** They're sealed into the signature, so a re-sign is the only moment they can change; offering one that silently kept the old numbers would send owners back through the side door for the other half of the job.

And because a cap edited and forgotten is the same class of surprise as one silently kept, the panel names what's about to change:

> Signing now also changes: per-trade 50 USDG → 10 USDG · daily 500 USDG → 50 USDG.

Derived from `grant.caps` rather than a form snapshot, so it stays true however the edit was made — including a preset click higher up the page.

When the browser holds no owner key, the section says so and points at the two things that do work, rather than rendering a button that returns immediately.

Both stale references now point at this control.

## Verification

`npm test` 1,408 passing · `npx tsc --noEmit -p web` clean · `npm run build` clean.

Exercised in a browser against a seeded throwaway grant (fake key material, cleared afterwards): the section renders in the loaded panel, the number inputs drive React state through the real `onChange`, and the change notice reads correctly. Screenshot behaviour confirmed at both mobile and desktop widths.

## Not covered

The button calls the existing `renewKey()` unchanged, so anything true of renewal is still true here — including that it re-signs on the **currently selected chain**, which follows the loaded grant. It cannot be used to move a key between chains; that is still "switch to another wallet", correctly.

### DIFF
--- a/web/src/app/grant/page.tsx
+++ b/web/src/app/grant/page.tsx
@@ -586,6 +586,30 @@ export default function GrantPage() {
     setFunding(null);
   }
 
+  /**
+   * Which caps the form now differs from the SIGNED ones, in words.
+   *
+   * The re-sign panel lets caps be edited in place, and an owner who tweaked a
+   * number and scrolled away should not discover the change by reading a
+   * receipt later. Derived from `grant.caps` rather than from a snapshot of the
+   * form, so it stays true no matter how the edit was made — including a preset
+   * click higher up the page.
+   */
+  const capChanges = grant
+    ? (
+        [
+          ["per-trade", caps.perTradeUsdg, grant.caps.perTradeUsdg, "USDG"],
+          ["daily", caps.dailyUsdg, grant.caps.dailyUsdg, "USDG"],
+          ["trades/day", caps.maxOpsPerDay, grant.caps.maxOpsPerDay, ""],
+          ["expiry", caps.expiryDays, grant.caps.expiryDays, "days"],
+          ["breaker", caps.maxDrawdownPct, grant.caps.maxDrawdownPct, "%"],
+        ] as const
+      )
+        .filter(([, now, was]) => now !== was)
+        .map(([label, now, was, unit]) => `${label} ${was}${unit && ` ${unit}`} → ${now}${unit && ` ${unit}`}`)
+    : [];
+  const capsEdited = capChanges.length > 0;
+
   const gasFunded = (funding?.gasWei ?? 0n) > 0n;
   const usdgFunded = (funding?.usdgUnits ?? 0n) > 0n;
   // Once a grant exists, the truth is what's IN it — not the selector state.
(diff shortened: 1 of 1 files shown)


millw14:
Extended to cover the second half of the same gap.

A new user on a testnet faucet asked **"how to do leave testnet? still understanding UI"** and couldn't find the answer, because there isn't one on the page. The chain picker renders only under `!grant || switching`, so the only route off testnet was a control labelled **"switch to another wallet"** — which is where the picker happens to live, and which sounds like it throws away the wallet you have.

That's worse than a mislabelled button: **testnet cannot trade.** Every token and router address merrymen knows is a mainnet-4663 deployment, so a testnet balance reads as zero and every route is refused (`preflight.ts` chain guard). A user on 46630 isn't in a safer version of the product — they're in one that cannot do the thing they came for, with no signposted exit.

The re-sign panel now carries the move as a named action: a checkbox that says what practice mode actually is, the real-funds warning and acknowledgement when moving toward mainnet, a button that renames itself to **"move to real money & re-sign"**, and a change line that leads with `chain practice → mainnet`.

**Deliberately not a selector.** The mount effect pins `chainId` to the loaded grant precisely so a mainnet owner cannot press renew and land on the sandbox — the comment there calls it load-bearing. A free dropdown here would reintroduce that drift. An explicit checkbox that resets the acknowledgement on every toggle keeps the property: the chain moves only when someone says so, twice.

The button is **disabled rather than hidden** while the acknowledgement is missing — a control that vanishes leaves the reader wondering what they did wrong.

Verified in a browser against a seeded testnet grant (that user's exact state): ticking the move flips the label, reveals the warning, disables the button, and prints `Signing now also changes: chain practice → mainnet`; ticking the acknowledgement re-enables it. Fake key material, cleared afterwards.

PR #32, opened 2026-08-31
The deploy script had a parse error, and nothing was checking it

millw14:
`scripts/deploy-ponsselftrade.ts` line 135 read:

```ts
writeFileSync(file, JSON.stringify(book, null, 2) + "
");
```

A literal newline inside a string literal — a parse error. I put it there in the commit that made the script record its own deployed address. **That script's entire job is to spend real gas deploying a contract to mainnet, and the owner was about to run it.**

## Why nothing caught it

`contracts/tsconfig.json` includes only `./hardhat.config.ts` and `./test`. Nothing under `scripts/` has ever been compiled — not locally, not in CI.

I ran `tsc --noEmit -p tsconfig.json` after writing it, grepped the output for the filename, saw zero lines, and reported the file clean. **The file was never in the program.** Zero errors in a file nobody compiled is not evidence of anything, and I offered it as though it were.

It was found by an architecture-mapping pass *reading* the file, not by a tool.

## The fix, and the hole

- The parse error is fixed.
- `tsconfig.scripts.json` now compiles `scripts/`, and a CI step runs it.
- Separate from the root config rather than widening it: these run under Hardhat's ESM loader and use `import.meta.url`, which `module: commonjs` rejects. Changing the root would alter how the tests and the Hardhat config compile, to fix a problem that belongs to one directory. It includes `hardhat.config.ts` because that registers the hardhat-viem type augmentation — without it every `hre.viem` reads as an error.
- `deploy-v4selfswap.ts` was equally unchecked and is now covered too.

---

Same escape-collapse failure as three others today (`/s+/g` eating every letter "s", `\b` becoming U+0008, `/$15/` matching nothing). Every one was a backslash written through a generator. The difference here is that the others failed loudly in tests, and this one sat in the only directory with no tests at all.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,6 +66,13 @@ jobs:
         working-directory: contracts
         run: npm install --no-audit --no-fund
 
+      # DEPLOY SCRIPTS ARE CHECKED TOO. The root tsconfig covers only the config
+      # and ./test, so scripts/ was never compiled by anything — and one shipped
+      # with a parse error, in a file whose job is to spend real gas on mainnet.
+      - name: Typecheck deploy scripts
+        working-directory: contracts
+        run: npm run typecheck:scripts
+
       - name: Test
         working-directory: contracts
         run: npm test
--- a/contracts/package.json
+++ b/contracts/package.json
@@ -5,7 +5,8 @@
   "description": "On-chain drawdown breaker: BreakerRegistry + Kernel v3 policy adapter.",
   "scripts": {
     "build": "hardhat compile",
-    "test": "hardhat test"
+    "test": "hardhat test",
+    "typecheck:scripts": "tsc --noEmit -p tsconfig.scripts.json"
   },
   "devDependencies": {
     "@nomicfoundation/hardhat-viem": "^2.0.0",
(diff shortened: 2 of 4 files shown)


millw14:
**Small and still valid — just needs a rebase.**

4 files, +33/−3, but CONFLICTING against current `main`.

**What to do:** rebase and confirm `app` is green. Since the point of this PR is that nothing was checking the deploy script, please also confirm the typecheck it adds actually fails on the original parse error — otherwise the guard lands without evidence it guards anything.

---

millw14 (commented):

**Verdict: The small code change is sound; resolve integration conflicts before merge.**

Reviewed the literal-newline parse fix, the separate deployment-script tsconfig, its module settings, and the CI invocation. The dedicated config includes Hardhat config for its type augmentation without changing the test compilation mode. No actionable code defect found in this patch. GitHub reported merge conflicts at review start, so update against main and retain equivalent script coverage through conflict resolution.

Validation: existing app/contracts checks and Vercel check passed for the reviewed snapshot. Deployment scripts were not executed (they can spend funds), and I did not rerun the contracts toolchain locally.

PR #56, opened 2026-09-03
The first operation was refused by our own doubling, not by the chain

millw14:
Twenty-four real swap attempts died on `AA21 didn't pay prefund`, and `boundGas` never received a number on any of them — so all twenty-four were logged as `gas-unreadable`, a refusal that names our own blindness and nothing else. Two independent causes, both proven experimentally before production code was touched.

## The diagnosis, proven first

`spikes/first-op-gas/probe.ts` builds a synthetic Kernel account with the identical wall shape and runs seven experiments against Pimlico on 4663:

| # | Question | Result |
|---|---|---|
| 1 | Does the real first op estimate? | `AA21 didn't pay prefund` |
| 2 | With a sender balance override? | Runs into execution (`STF`) — the estimate exists |
| 3 | Control: override an unrelated address | `AA21` again → **the override is genuinely applied** |
| 4 | Does the estimate move with override size? | Byte-identical at 0.01 / 0.05 / 0.2 / 1 ETH → **not a search artefact** |
| 5 | Decompose: sudo-only vs walled | 460,797 → 7,711,654 raw. **The wall is +7,059,814 verification.** |
| 6 | Per-permission slope (`small-wall.ts`) | 2→3,573,604 … 18→15,423,182. ~370,299 gas/permission, base ~1.05M |
| 7 | Does the nonce encode the enable (`nonce-mode.ts`)? | walled `0x0102630974640…` vs sudo-only `0x0000845adb2c…` |

Two claims from earlier reports were **refuted by this work and are withdrawn**: the 18.6M gas figure was never real (`gas-absurd` occurs **0 times across all 14 deployments** — `boundGas` was never handed a number to judge), and "the account is underfunded for the simulation" is wrong, per experiment 4.

## Cause 1 — the estimate could not run

viem sends `eth_estimateUserOperationGas` with no gas limits, so the bundler substitutes its own for the simulation, and the EntryPoint's prefund check then runs against the account's **real** balance against limits nobody chose. On 4663 the block gas limit is `1.125e15`. No balance a person would fund an agent with survives that multiplication.

```ts
stateOverride: [
  { address: account.address, balance: bounds.absoluteMax * feeCeiling * 2n },
],
```

**A state override is a parameter of one RPC call.** It is not a field of a UserOperation, so it cannot be signed, hashed or broadcast. It does not touch the account's balance, the prefund the EntryPoint actually charges, any policy check, the calldata, or any on-chain state. It is sized from *our own ceiling* rather than a round number, so the simulation is never handed more room than we would ever sign for.

And because a simulation that can pay says nothing about an account that can, the **real** balance is now read after the estimate and before anything is signed:

```
gas refused (prefund-short): this operation needs 9677201 gas, which the EntryPoint
requires the account to HOLD as … wei at … wei/gas. It holds … wei — short by ….
Nothing was signed. The account is charged only for gas USED, but it must hold the
full amount up front, so this is a funding shortfall and not a refusal by the chain.
```

Without that check this patch would have converted an opaque AA21 at estimation into an opaque AA21 at submission, having spent a signature on the way.

## Cause 2 — the doubling was the refusal

The measured first operation is `verification 7,418,031 · preVerification 243,443 · call 50,180`. Under a blanket 2× that signs **15,423,308** — while the **raw** total, 7,711,654, clears the 8,000,000 ceiling it was being refused by. The headroom, not the ceiling, was the entire refusal.

The reasoning behind the blanket 2× — *"too high costs a slightly larger prefund, too low costs the entire operation"* — is derived from `callGasLimit` and is false for the other two:

| Field | Too low ⇒ | Cost | Headroom |
|---|---|---|---|
| `callGasLimit` | call OOGs, `success=false` | **account pays in full** | **×2** |
| `verificationGasLimit` | validation reverts, `FailedOp`, never bundled | nothing | **×1.25** |
| `preVerificationGas` | bundler rejects before inclusion | nothing | **×1.25** |

Two of three fail free and loud — the outcome this file already chooses deliberately everywhere else. Verification is deterministic given fixed calldata against a fixed account (signature check, CREATE2 of fixed bytecode, zero-to-nonzero SSTOREs; the only drift is warm storage, which is *cheaper*). `gasUsedForL1` is 0 on 4663, so nothing moves under `preVerificationGas` either.

## The paymaster hole

`totalGas` counted three fields. **The EntryPoint's prefund counts five**, and `paymaster.ts` allows up to 500,000 in each of the other two — so up to **a million gas of prefund sat outside every bound in this file**. A ceiling that cannot see a fifth of what it is bounding is not a ceiling.

They are counted now, and **never multiplied**: they are the sponsor's numbers, bounded by `paymaster.ts` and asserted unchanged by `assertBoundsHeld`, and inflating somebody else's limits is not ours to do. A paymaster appearing on a **self-paying** operation is refused outright as the new `gas-paymaster-unexpected` rather than passing unseen — the "or explicitly prove they are absent" half of the requirement.

## The first-enable ceiling, derived rather than chosen

```
  verificationGasLimit   7,418,031 × 1.25 =  9,272,538
  preVerificationGas       243,443 × 1.25 =    304,303
  callGasLimit              50,180 × 2.00 =    100,360
                                   bounded =  9,677,201

  safety margin, documented separately rather than folded in:
    a tenant's own custom tokens widen both ONE_OF lists, at a measured
    ~370,299 raw / ~462,874 bounded each. The margin buys FIVE.
    9,677,201 + (5 × 462,874)          = 11,991,571
                                        ->  12,000,000
```

1.24× the base operation. Deliberately not generous: at ~16 custom tokens the estimate exceeds it and the op is refused, which is the kinder failure given the cliff below.

**`GAS_BOUNDS.absoluteMax` is still 3,000,000.** The ordinary ceiling does not move, and a test pins that the measured wall offered ordinary bounds is still refused.

`DEPLOY_GAS_BOUNDS` is **removed** rather than left dead — nothing read it after this change, and an 8M constant sitting beside a 12M one, both claiming to be the first-op ceiling, is the shape of the next bug.

## Fail closed on execution state

"The account has no code" is a fact about an *address*, and admits every shape an undeployed account could send. The wide ceiling now requires the **operation** to prove, out of its own nonce, that it carries a permission-validator enable:

```ts
const firstEnable = !accountLive && isFirstEnable(nonce);
const bounds = firstEnable ? FIRST_ENABLE_GAS_BOUNDS : GAS_BOUNDS;
```

Kernel v3 packs `mode(1) ‖ vType(1) ‖ validator(20) ‖ id(2) ‖ seq(8)`. Both bytes are required — mode alone would also admit an enable of some *other* validator type, which is not the operation this ceiling was measured for. Undeployed alone buys nothing, which also means the memoised `isDeployed` read failing open can no longer widen anything by itself.

## Tests

All ten requested, in `worker/src/gas-limits.test.ts`:

1. `stateOverride` appears exactly once, inside the estimate call's own argument object, and nowhere after `sendUserOperation`
2. the override value is not hoisted into a variable that could travel; `feeCeiling` and `bounds.absoluteMax` never appear in the send
3. `callGasLimit` gets 2×
4. verification and preVerification get 1.25× — pinned as a comparison too, so an edit setting all three equal fails
5. the ordinary ceiling did not move; the measured wall under `GAS_BOUNDS` is still `gas-absurd`
6. `isFirstEnable` needs both bytes: sudo-only, enable-of-sudo, and already-installed-permission all refused; sequence bytes ignored
7. an estimate past 12M is still `gas-absurd` under the first-enable bounds
8. paymaster gas cannot walk under the ceiling — 2.4M alone passes, 3.4M with the sponsor's million refuses
9. `prefund-short` is priced from the bounded total, happens before the send, and says how short
10. **the measured 18-permission wall passes**, at exactly 9,677,201 — and the same estimate under the old blanket 2× is still refused, so the per-field headroom did the work rather than the ceiling alone

Plus a test that recomputes every term of the ceiling's arithmetic, and asserts it does not exceed its own stated margin by another token's worth.

`wall.test.ts` and `paymaster.test.ts` pass **untouched**. `npm test` — 1911 tests, 0 failures. `npm run typecheck` clean across worker, web and browser.

## Deliberately not here

- **Option B** (moving plugin installation into the owner/relay flow) — it would put `delegatecall` into a path that categorically refuses it, and create installed-validator lifecycle state that complicates renewal and revocation.
- **Option C** (shrinking the 18-permission wall to make gas cheaper) — product scope.
- **No real trade is broadcast by this PR.**
- The **`usableExtraTokens` cliff is real**: past roughly 40 custom tokens a grant provably cannot validate at all (`AA23 reverted 0x`, measured). That gets its own issue — a maximum supported permission/token count or pre-signing size validation — rather than a silent truncation of somebody's grant.

### DIFF
--- a/spikes/first-op-gas/check-extras.ts
+++ b/spikes/first-op-gas/check-extras.ts
@@ -1,0 +1,99 @@
+/**
+ * Do `extraTokens` actually widen the wall's policies, and is the estimation
+ * stub the REAL enable blob or a fixed-size dummy?
+ *
+ * The gas sweep returned an identical 7,418,031 verificationGasLimit at +0, +5,
+ * +15 and +40 extra tokens, with the stub pinned at 10,932 bytes. Either the
+ * wall's size genuinely does not drive the cost, or the extras never reached
+ * `buildWallPolicies`, or the stub is a constant that does not reflect the real
+ * enable data. Those three have completely different consequences, so this
+ * separates them before any conclusion is drawn.
+ *
+ * No network writes. Nothing signed, nothing broadcast.
+ */
+import { createPublicClient, http, type Address } from "viem";
+import { generatePrivateKey, privateKeyToAccount } from "viem/accounts";
+import { getEntryPoint, KERNEL_V3_3 } from "@zerodev/sdk/constants";
+import { createKernelAccount } from "@zerodev/sdk";
+import { signerToEcdsaValidator } from "@zerodev/ecdsa-validator";
+import { toPermissionValidator } from "@zerodev/permissions";
+import { toECDSASigner } from "@zerodev/permissions/signers";
+import { buildWallPolicies, chainForId } from "../../packages/core/src/index";
+
+const CAPS = { perTradeUsdg: 50, dailyUsdg: 500, expiryDays: 14, maxDrawdownPct: 10, maxOpsPerDay: 48 };
+
+async function main() {
+  const publicClient = createPublicClient({
+    chain: chainForId(4663),
+    transport: http("https://rpc.mainnet.chain.robinhood.com"),
+  });
+  const entryPoint = getEntryPoint("0.7");
+
+  for (const n of [0, 40]) {
+    const extras = Array.from({ length: n }, (_, i) => ({
+      // isValidCustomToken requires exactly {symbol, address, decimals} — my
+      // first attempt passed the STOCK_TOKENS shape (name/chainlinkFeed/kind,
+      // no decimals) and every entry was silently dropped, which made the sweep
+      // look like the wall size did not matter when it had never been varied.
+      symbol: `X${i}`,
+      address: `0x${(i + 0x1000).toString(16).padStart(40, "0")}` as Address,
+      decimals: 18,
+    }));
+    const owner = privateKeyToAccount(generatePrivateKey());
+    const ecdsa = await signerToEcdsaValidator(publicClient, {
+      signer: owner,
+      entryPoint,
+      kernelVersion: KERNEL_V3_3,
+    });
+    const sudoOnly = await createKernelAccount(publicClient, {
+      entryPoint,
+      kernelVersion: KERNEL_V3_3,
+      plugins: { sudo: ecdsa },
+    });
+    const { policies } = buildWallPolicies({
+      caps: CAPS,
+      smartAccount: sudoOnly.address,
+      ...(n ? { extraTokens: extras as never } : {}),
+    });
+
+    // The CALL policy's own serialized data is where the ONE_OF lists live.
+    const sizes = policies.map((p: unknown) => {
+      const o = p as { policyParams?: { type?: string }; getPolicyData?: () => string };
+      let bytes: number | string = "n/a";
+      try {
+        const d = o.getPolicyData?.();
+        if (typeof d === "string") bytes = d.length / 2 - 1;
+      } catch {
+        bytes = "threw";
+      }
+      return `${o.policyParams?.type ?? "?"}:${bytes}`;
+    });
+
+    const sess = await toECDSASigner({ signer: privateKeyToAccount(generatePrivateKey()) });
+    const pv = await toPermissionValidator(publicClient, {
+      entryPoint,
+      kernelVersion: KERNEL_V3_3,
+      signer: sess,
+      policies,
+      flag: "0x0002",
+    });
+
+    let enableBytes: number | string = "n/a";
+    try {
+      const v = pv as unknown as { getEnableData?: (a: Address) => Promise<string> };
+      const e = await v.getEnableData?.(sudoOnly.address);
+      if (typeof e === "string") enableBytes = e.length / 2 - 1;
+    } catch (err) {
+      enableBytes = `threw: ${String(err).slice(0, 40)}`;
+    }
+
+    console.log(
+      `extras ${String(n).padStart(2)} · policies ${policies.length} · [${sizes.join(" ")}] · enableData ${enableBytes}B`,
+    );
+  }
+}
+
+main().catch((e) => {
+  console.error(String(e).slice(0, 400));
+  process.exit(1);
+});
--- a/spikes/first-op-gas/gate-v2-verify.mjs
+++ b/spikes/first-op-gas/gate-v2-verify.mjs
@@ -1,0 +1,116 @@
+/**
+ * TWO CLAIMS THE GATE RESTS ON, CHECKED WITHOUT THE SDK.
+ *
+ * READ-ONLY: eth_call only (EntryPoint.getNonce and Kernel.permissionConfig).
+ *
+ * CLAIM 1 (C2's premise). After a permission enable LANDS, the EntryPoint's
+ *   counter for that ENABLE key is 1 and never 0 again, while the DEFAULT key
+ *   for the same permission id carries the ordinary traffic. If that holds,
+ *   `sequence == 0` is a genuine once-per-inclusion bound and refuses only
+ *   operations that would revert.
+ *
+ * CLAIM 2 (C4's premise). permissionConfig(pId) read straight off the account
+ *   distinguishes installed from not-installed per (account, permissionId), and
+ *   a codeless account answers with empty returndata rather than with zeros —
+ *   which is why "no code" must be a separate branch and not a parse result.
+ */
+const RPC = "https://rpc.mainnet.chain.robinhood.com";
+const EP = "0x0000000071727De22E5E9d8BAf0edAc6f37da032";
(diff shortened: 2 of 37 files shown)


millw14:
## Review caught a real bug in this PR, and it is now fixed here

**The premise was wrong.** `!accountLive && isFirstEnable(nonce)` assumed a permission-validator enable is one-time per **account**. It is one-time per **session key**. Kernel keys the install on `(account, permissionId)`, a renewed key is a new `permissionId`, and the account it installs into has been deployed since the first arm. Every one-click renewal would have been routed through the 3,000,000 ceiling and refused.

### Measured, three independent ways (Pimlico, chain 4663, 2026-09-03)

| what | raw | bounded |
|---|---|---|
| `0xa48cE91e…` real deployed account, fresh wall | 7,401,680 | 9,289,735 |
| `0x032Da6A0…` merrymen's own live account | 7,530,220 | 9,450,410 |
| synthetic (cloned code + storage, no factory) | 7,569,825 | 9,499,915 |
| *undeployed first op, for comparison* | *7,711,654* | *9,677,201* |

Deployment is worth **141,823–309,974 gas — 2–4%**. A bundler-free oracle (`eth_estimateGas` of `validateUserOp` alone) puts executed verification at 5,455,289, which is 1.82× the 3M ceiling before any headroom.

Controls that had to pass first: the same accounts sudo-only estimate at 190,426 / 220,075 with mode `0x00` nonces; the harness reproduces the undeployed 7,711,654 to within 63 gas; and **without** the owner-slot override every measurement is `AA23 reverted 0xc48cf8ee`, so the enable signature is genuinely validated and the numbers are not an artefact of a skipped check.

**Shape half:** 28 rows across 4 real deployed accounts × 7 wall/key variations — every one `mode 0x01 / vType 0x02` with `accountLive` true. Zero counterexamples. The converse holds: 4 accounts whose validator *did* land drop to `mode 0x00`, so the steady state keeps its 3M ceiling and dropping the deploy-state conjunct does not touch it.

### The gate now asks the operation and lets the chain answer

`readEnableState` requires all three:

1. the nonce is enable-shaped (as before)
2. **its sequence is 0** — the EntryPoint increments only on *inclusion*, so a rejected or retried enable stays at 0 while a landed one reads 1 forever. Verified on chain. Nothing local chooses these bytes, so the wide ceiling admits at most **one included operation per (account, permissionId)**, with no ledger to go stale.
3. **`permissionConfig(id)` on the account is empty** — verified keyed by `(account, id)`: `0x3ca1cec8` reads INSTALLED on one live account and absent on another; a codeless address answers with empty returndata.

**Condition 3 is not belt-and-braces.** @zerodev's `isPluginEnabled` catches *every* read error to `false` — both disjuncts — so one flaky `eth_call` makes the SDK build an ENABLE-shaped op for an already-installed validator. Fault-injected and confirmed: healthy read → `0x00`, one broken read → `0x01`, same account, same id. On the nonce alone that op would have been handed 12,000,000 for work nobody needs. Asking the chain ourselves converts the SDK's fail-open into our fail-closed.

And it **fails closed, inverting** the rule `isDeployed` uses twenty lines above: there a failed read answers "not deployed" because being wrong that way only narrowed a ceiling; here "no code" is a reason to *widen*, so a read we could not make must never manufacture one.

Three named refusals, because falling through to the ordinary ceiling would refuse these too — as `gas-unreadable`, which says only that we failed to get a number: `enable-replayed`, `enable-redundant`, `enable-unverified`.

### What did not change

**12,000,000 stands.** A renewal is the same enable minus initCode and CREATE2, so it is strictly cheaper — the dearest measures 9,499,915 bounded against the 9,677,201 the constant was derived from. The undeployed first arm remains the binding case and the five-token margin is intact. `GAS_BOUNDS.absoluteMax` is still 3,000,000. `isDeployed` survives, demoted from guard to **label** on the `[gas]` line — a test pins that every *read* of `accountLive` is inside that line.

Also corrected: the constant's comment claimed the ceiling admits "~16 custom tokens". Measured at ~512 blob bytes and ~358,884 raw gas each, it is refused at **six** — matching the five-token margin two paragraphs above. The prose was wrong by 3×.

### Verification

Four mutations, each caught by exactly one test: reinstating `!accountLive`; dropping the sequence check; failing open on an unreadable `permissionConfig`; and the `>> 224` off-by-two on the permission id — worth naming, because it returns the mode and type bytes glued to half the id, so every lookup answers "absent" and the check silently becomes one that always passes.

**1918 tests, 0 failures.** Typecheck clean across worker/web/browser. `wall.test.ts` and `paymaster.test.ts` untouched. Nothing broadcast, no funds moved, no grant re-signed.

millw14:
Superseded by #60, which carries these exact gas-correctness commits with the Stage B RPC and reconciliation work split out onto `stage-b/rpc-and-reconciler` so the first-trade fix could merge unbundled. #60 is merged.

PR #59, opened 2026-09-03
feat: fluid trading — say it, confirm it, it lands

aor-rex:
## Summary

19 commits that turn merrymen from a "type exact slash commands" bot into a fluid trading agent. The bot now accepts natural trade intents by symbol or contract address, parks every trade for confirmation with inline buttons, and routes through the correct venue (Uniswap pools, Pons bonding curves, or on-chain curve lookup).

**Live-tested:** 3 real basket buys landed on chain 4663 (NVDA, TSLA), 3 auto-conversions landed, account deployed and verified on-chain.

### Trading fluidity
- Fast parse: `buy 0x... 1`, `$NVDA 1usdg`, `/buy PIPOTAM 5` all work — typo-tolerant, strips `$` and `usdg` suffix, supports 10-char memecoin names
- Every trade parks for `/confirm` with inline ✅/❌ buttons — nothing fires until you confirm
- Ask-amount context: "how much?" → user says "5" → trade parks instantly, no LLM call
- Trade outcome surfaced after confirm: "✅ landed", "🚫 rejected (reason)", "❌ reverted"
- Gas warning on the parked card when the wallet is low on ETH

### Pons memecoin support
- `*pons` wildcard: one re-sign unlocks every launchpad token — no need to add each one individually
- On-chain Pons factory verification (`verifyPonsToken`) — confirms tokens by raw query against the factory
- On-chain curve lookup (`curveForToken`) — finds the curve address from the factory event for tokens not in the local cache
- Caches discovered curves so subsequent calls use the local DB

### Policy fixes
- Drawdown breaker no longer blocks user-initiated chat trades — the owner knows what they're doing
- Reconcile scan timeout (90s) — prevents 30-minute startup stalls under rate-limited RPC (observed live)
- Clean AI-down fallback — no more thinking dumps, replaced with `/buy SYMBOL AMOUNT` signpost

### Parser improvements
- `parseSlash` accepts contract addresses as buy/sell targets with `address` field
- `fastParseTrade` skips symbol resolution when address is provided
- `Command` type extended with optional `address` field
- ask-amount type extended with `address` for address-based follow-ups

## Testing
- 18 fluid-trading tests pass
- 52 interpreter tests pass
- All policy tests pass
- Full sandbox: 12/12 core flows passing
- Worker typecheck: clean

## Still needed (one-time setup for Pons tokens)

The `PonsSelfTrade` adapter contract needs to be deployed on chain 4663 (one-time, ~$2-5 gas), the address pasted in `/settings`, and the grant re-signed. See `docs/owner-runbook-pons.md` for instructions. The code is ready; the deploy step is owner action.

### DIFF
--- a/packages/core/src/grant.ts
+++ b/packages/core/src/grant.ts
@@ -344,6 +344,11 @@ export function builtinGrantTargets(grant?: Pick<StoredGrant, "grantFeatures"> |
 export function sellableAssets(grant: Pick<StoredGrant, "grantFeatures" | "grantTokens"> | null): Set<string> {
   const set = builtinGrantTargets(grant);
   for (const a of grant?.grantTokens ?? []) set.add(a.toLowerCase());
+  // WILDCARD: "*pons" in grantTokens means every token discovered through the
+  // Pons launchpad is treated as approved. The actual token list is checked at
+  // policy time against knownCurves/ponsAssets. The sentinel value is always
+  // lowercase so a simple includes check works.
+  if (grant?.grantTokens?.includes("*pons")) set.add("*pons");
   return set;
 }
 
--- a/web/src/app/grant/page.tsx
+++ b/web/src/app/grant/page.tsx
@@ -413,6 +413,7 @@ export default function GrantPage() {
   // the wall at signing — which is why it is read here and not at trade time.
   const [v4Adapter, setV4Adapter] = useState<`0x${string}` | undefined>(undefined);
   const [ponsAdapter, setPonsAdapter] = useState<`0x${string}` | undefined>(undefined);
+  const [ponsWildcard, setPonsWildcard] = useState(false);
   // The basket matters here for the same reason: /settings offers every registry
   // symbol, but only the ones sealed into the signature can be sold.
   const [basketSymbols, setBasketSymbols] = useState<string[]>([]);
@@ -563,6 +564,7 @@ export default function GrantPage() {
         v4AdapterAddress: v4Adapter,
         ponsAdapterAddress: await verifiedAdapter(ponsAdapter, chainId, setStatus),
         hostedAs: session.hosted ? (session.address ?? undefined) : undefined,
+        ponsWildcard,
       });
       setGrant(g);
       // Take the ARMED state from what the server actually said. This used to be
@@ -614,6 +616,7 @@ export default function GrantPage() {
         v4AdapterAddress: v4Adapter,
         ponsAdapterAddress: await verifiedAdapter(ponsAdapter, chainId, setStatus),
         hostedAs: session?.hosted ? (session.address ?? undefined) : undefined,
+        ponsWildcard,
       });
       // They just pasted the owner key, so it's demonstrably backed up — skip the
       // backup gate and drop them straight into the funded/manage view.
@@ -1163,6 +1166,20 @@ export default function GrantPage() {
                 It rested on ZeroDev's rate-limit policy, whose contract has no bytecode on
                 Robinhood Chain.
               */}
+            </div>
+
+            <div className="caps-section" style={{ marginTop: 16 }}>
+              <label className="ack-row">
+                <input
+                  type="checkbox"
+                  checked={ponsWildcard}
+                  onChange={(e) => setPonsWildcard(e.target.checked)}
+                />
+                <span>
+                  <b>Allow all Pons launchpad tokens</b> — trade any memecoin from the Pons feed
+                  without adding it in /settings. Each trade still respects your caps above.
+                </span>
+              </label>
             </div>
 
             {mode === "create" ? (
--- a/web/src/lib/session.ts
+++ b/web/src/lib/session.ts
@@ -189,6 +189,11 @@ async function mintGrant(
    * client can trust.
    */
   hostedAs?: Address,
+/**
+   * Opt-in: include "*pons" in grantTokens so every Pons launchpad token
+   * passes the asset-allowlist check. No on-chain permission change.
+   */
+  ponsWildcard?: boolean,
 ): Promise<MintedGrant> {
   // Testnet is the sandbox; mainnet (4663) is real funds — the UI gates that
   // choice behind an explicit consent step. Note: the call-policy addresses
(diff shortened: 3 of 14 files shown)


millw14:
Reviewed in depth. The confirmation-card idea underneath this is good and I want it — parking every trade behind an explicit ✅ is the right shape. But I can't merge it as it stands: eight blockers, and the headline one means the `*pons` feature cannot work at all in production.

## The wildcard is signed off-chain only

`web/src/lib/session.ts:336` adds `*pons` to the stored grant's token list, and `worker/src/policy.ts` learns to pass it. Nothing adds it to the **on-chain permission wall** — `buildCallPermissions` still seals a `ONE_OF` list of literal addresses, and `*pons` is not an address.

So the software gate says yes, the UserOp is built and signed, and the session key's validation refuses it. Every wildcard buy fails at the wall, after we've spent the gas getting there. The "one re-sign unlocks every launchpad token" line in the description can't hold — a signature can only cover addresses it named.

This needs a design decision before code: either the adapter is the thing that's allowlisted (pin the curve adapter and let it fan out, which the grant already supports via `ponsAdapterAddress`), or each token really does need naming. There isn't a third option that a signed policy can express.

## `knownPonsAssets()` is not the Pons set

`worker/src/store.ts:2540` returns every token the discovery feed has ever written. That feed is an untrusted external index, not the launchpad. So the wildcard's allowlist is "anything an external service told us about", which is a much larger and much less trustworthy set than "launched by the Pons factory".

## The drawdown bypass isn't conditioned on anything

`worker/src/policy.ts:571` is fine in itself, but every call site passes `true` unconditionally (`index.ts:4906, 4951, 4976, 5030, 5054`). There's no check that the message came from the owner, no check that it's interactive — the flag is just always on for chat. The description says "the owner knows what they're doing", and I agree with that as a principle; the code doesn't implement it, it implements "chat is exempt".

Note also it's applied to `transfer`, which isn't a trade.

## `curveForToken` decodes both fields wrongly

`worker/src/venues/pons.ts:90`:

```ts
const curve = first.topics[2] as `0x${string}` | undefined;
const quoteToken = first.data ? (`0x${first.data.slice(2, 42)}`) as `0x${string}` : undefined;
```

`topics[2]` is a 32-byte topic — an indexed address is left-padded, so this returns a 66-character string where an address is expected. And `data.slice(2, 42)` takes the first 20 bytes of the first word, which is the **padding**; the address is in the last 20.

The correct decoders are 80 lines below in the same file:

```ts
curve: addressFromTopic(log.topics[2]!),
quoteToken: `0x${log.data.slice(2 + 24, 2 + 64)}`.toLowerCase() as `0x${string}`,
```

Both wrong values are then written into the table the curve-provenance rule reads.

The PR also deletes the comment that documented the event shape ("(token indexed, curve indexed, creator indexed) plus THREE data words… 11,395 launches sampled in 24h"). That comment was the specification the new code needed.

## `verifyPonsToken` verifies less than it sounds like

It checks that the token appears in a factory launch event. That's "was deployed via Pons" — anyone can deploy via a launchpad, so it isn't vetting. The comment in `policy.ts` says "The launchpad pre-vets the token", and that claim is doing the load-bearing work for the whole widening.

It also returns `false` on a thrown RPC call, so "not a Pons token" and "the chain didn't answer" are the same answer. In a policy check that has to fail closed, those need to be different.

## It adds an unbounded `getLogs` to every swap

`worker/src/index.ts:2592` calls `verifyPonsToken` on the intent path, which does a 300,000-block `getLogs` — on **every** swap intent, including strategy trades that have nothing to do with Pons. That's the reconcile-stall failure mode this PR is separately fixing, reintroduced somewhere else.

## The parser can turn a sentence into a parked trade

`worker/src/telegram/service.ts:609` — the typo table maps `but`, `guy`, `bug` and `seal` to trade verbs, and the LLM symbol resolver's output isn't validated against the tradable universe. "but I'm not sure" is one token away from a parked buy.

---

Majors worth fixing in the same pass: `no-exit` is bypassed by the wildcard (`policy.ts:410`), re-opening the one-way door its comment describes; the inline ✅ carries no reference to the card it's attached to, so tapping an old card fires the newest parked trade (`service.ts:780`); the ask-amount follow-up branch is unreachable dead code, so "how much? → 5 → parks instantly" never happens (`service.ts:736`); chat trade replies read a never-reset global and can report the previous trade's outcome as this one's (`index.ts:4952`); and the grant-page caption claims caps that no longer hold (`web/src/app/grant/page.tsx:1179`).

Happy to look again once the wildcard question is settled — that one's architectural and worth talking through before you write more code against it.

millw14:
**Needs a rebase, a sequencing decision, and an explicit answer on what gates it.**

CONFLICTING against current `main`, 21 commits, and the CI result here predates a lot of movement.

**It overlaps three other open PRs.** `packages/core/src/grant.ts`, `web/src/lib/session.ts`, `worker/src/index.ts`, `worker/src/policy.ts` and `worker/src/store.ts` are also touched by #38 / #101 / #102, and `telegram/interpreter.ts`, `executor.ts` and `service.ts` are also touched by #107. `telegram/api.ts` and `service.ts` moved again when #10 merged today. Rebasing all of these independently will just move the conflicts between them.

**One file is gone:** `web/src/app/grant/page.tsx` was removed by the terminal rewrite (#96, #103); everything else it modifies still exists.

**The part that needs a written answer.** This PR makes a trade land from a chat message, and it edits `worker/src/limits.ts`, `worker/src/policy.ts` and `worker/src/venues/pons.ts` — the permission and launchpad path. Please state in the description exactly which limits a fluid trade is checked against, and confirm it cannot bypass the per-trade and daily caps that a session key is bound by. That is the claim worth reviewing here; the conflicts are the easy part.

**What to do:** agree the ordering with #38/#101/#102 and #107 first, then rebase onto current `main` and let CI re-run.

---

millw14 (commented):

**[P2] The amount follow-up branch is unreachable after classification** — `worker/src/telegram/service.ts:736`. The earlier `if (!cmd)` block always assigns a command through the LLM, its failure fallback, or the no-provider chat fallback. Consequently `if (!cmd && askCtx)` cannot consume a bare `5` after the bot asks how much to buy. The advertised deterministic follow-up instead enters chat/provider handling. Resolve a valid pending amount context before general LLM classification, retain sender authorization, and add a service-level test with the provider both available and unavailable.

Validation: 18 fluid-trading tests passed. They do not establish this service-level conversation. GitHub reported merge conflicts at review start and the snapshot has no app/contracts CI check. The deliberate drawdown bypass is also a policy change requiring explicit product acceptance against the existing risk-limit requirement; an ordinary trade confirmation should not silently stand in for changing that policy.

PR #63, opened 2026-09-04
 Reasoning belongs in its own bank 

aor-rex:
Reasoning models dump chain-of-thought into the chat (`Here's a thinking process…`, `The user is asking...`, `Draft: "I`). `maxTokens 500` gets spent on reasoning, `content` stays empty, Telegram shows the dump or truncates to `Draft: "I`.

Universal, not prefix list: reasoning is a separate provider channel (`reasoning_content` / `reasoning` on OpenAI-compatible, `thinking` on Anthropic). `llm.ts` asks reasoning models for `reasoning_effort none` + `include_reasoning false` and discards what still arrives. `interpreter.ts:stripThinkingBlock` is now `<think>` fallback only.

Soul stays grounded without recitation: `CHAT_SYSTEM` 1-4 sentences, warm, grounded ONLY in STATE. Narrator gets `narratorIdentityBlock` (name+stage+tone, no born/age/counts) in system, volatile recall/live in user STATE. `who are you` stays deterministic `soulInfo`. History scrubs old verbatim headers so they stop teaching new ones. `what can you do` is fluid soul voice (only `/help` slash is HELP_TEXT).

## What this does NOT do
No change to slash/transfer gates, maxTokens defaults, or wall. Anthropic already `thinking: disabled`.

## Verification
Typecheck clean, tests 1958/1958. `hi` warm without preamble, `who are you` deterministic soul, `5` bare-amount bypass, no `Draft: "I`.

### DIFF
--- a/worker/src/llm.ts
+++ b/worker/src/llm.ts
@@ -144,6 +144,8 @@ export async function llmToolCall(
   }
 
   // openai-compatible function calling (Groq, OpenAI, Gemini, xAI, DeepSeek, …)
+  // Some reasoning models (gpt-oss-120b, deepseek-r1, qwen3-thinking, nemotron) dump chain-of-thought
+  // into `content` or `reasoning_content`. We ignore that side-channel and only use tool_calls.
   const body = {
     model: creds.model,
     max_tokens: opts.maxTokens ?? 1024,
@@ -306,6 +308,22 @@ async function providerError(creds: LlmCreds, r: Response): Promise<string> {
   return `${creds.provider} ${r.status}${safe ? ` — ${safe.slice(0, 300)}` : ""}`;
 }
 
+/**
+ * Reasoning models (nemotron, deepseek-r1, qwen3-thinking, gpt-oss) may return
+ * chain-of-thought in `reasoning_content`, `reasoning`, or inline `<think>…</think>`
+ * blocks inside `content`. Strip that side-channel before returning to Telegram.
+ */
+function stripReasoningFromContent(content: string, reasoning?: string): string {
+  let out = content ?? "";
+  // reasoning_content/reasoning is a separate field — never append it; it's thinking.
+  // If content is empty and only reasoning exists, treat as no answer (caller throws).
+  if (!out.trim() && reasoning) return "";
+  // Remove <think>…</think> and <|think|>…<|/think|> blocks (multiline, case-insensitive)
+  out = out.replace(/<\|?think\|?>([\s\S]*?)<\/\|?think\|?>/gi, "");
+  out = out.replace(/<think>([\s\S]*?)<\/think>/gi, "");
+  return out;
+}
+
 export async function llmText(
   creds: LlmCreds,
   opts: { system: string; prompt: string; maxTokens?: number },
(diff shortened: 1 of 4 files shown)


millw14:
Merged, with two changes. Thank you for this — the reasoning dumps were a real and visible problem and the diagnosis is right.

**Kept whole:** the response-side fix. Reading the answer out of `content` and discarding `reasoning_content` / `reasoning` is the half that needs no model list and can't go stale, and it's the half that actually solves it.

**Removed: `askAmountCtx`.** It let a message consisting of nothing but `5` become a live buy, with no confirmation step — while the sibling `transfer` case parks and waits for `/confirm`. Three things compounded:

- It was primed by regex-matching the bot's **own outbound prose** (`service.ts:693`), and the nudge this PR adds for a bare number matches that regex. So: send "5" with nothing pending → bot emits the nudge → the nudge primes the context → send "5" again → trade.
- The ticker came from the first 1–6 letter word of an **earlier** message (`users[length - 2]`), and history already contains the current message — so `"QQQ looking strong"` → `"buy NVDA"` → `"5"` buys **QQQ**.
- The side was `/sell/i.test(text) ? "sell" : "buy"`, so "should I sell before I buy more?" primes a sell.

No expiry, keyed by `chatId` rather than `fromId`. The caps and the wall still bound it, but a cap is a ceiling, not consent.

The problem it was solving survives without it: a bare number still never reaches the classifier, because the deterministic nudge answers it for zero tokens. If you want the follow-up properly, it belongs in its own PR — park it through `deps.setPending` with `CONFIRM_TTL_SEC`, key it by `fromId`, re-display side/symbol/amount on the confirm, and carry them from the `Command` the classifier actually emitted rather than re-deriving from prose. That's the `transfer` pattern and it's the right one.

**Corrected on the way in:**

- `extra_body` isn't an OpenAI wire field — it's a python-SDK convenience the SDK unwraps before sending. As JSON it's an unknown key, so `include_reasoning: false` was never sent to anyone. Worse, it was going into the **classifier's** request too, one strict-schema provider away from a 400 that takes the whole natural-language surface down. Now sent at the top level, from a single `quietReasoning()` helper instead of four copies of the model list.
- `stripThinkingBlock` missed the case that actually happens: a model that runs out of budget mid-thought emits `<think>` with no closing tag, and the paired rules match nothing. It also needs both spellings of both tags — matching one form's opener against the other's closer treats a terminated block as unterminated and eats the answer behind it.
- The send path failed **open**: `strippedReply || reply` meant a reply that was entirely thinking stripped to `""` and the raw thinking went out instead, in exactly the case the strip exists for.
- Capability questions were rerouted from `HELP_TEXT` to the model's own voice. What this agent can do depends on the deployment's settings and only `HELP_TEXT` knows it, so an answer written from memory promises powers the owner may not have enabled.
- The history scrub was pinned to one deployment's agent name, so everywhere else it scrubbed nothing while claiming to — it's built from `getName()` now, escaped. Its `|| t.content` tail restored the text it had just stripped.
- The dedupe added for the removed bare-amount branch was silently dropping a legitimately repeated message.

Added 15 tests, including source-reading invariants that fail if the bare-number trade path, the prose-derived ticker, or the fail-open send ever come back.

PR #68, opened 2026-09-04
Gate the equity chart, and correct the dry run

millw14:
Two fixes found by looking at what the last change actually rendered and printed.

## The equity chart was publishing a return the same page refused to publish

The canary's profile said **"capital on record is not evidenced"** and declined to show a return. Three lines below it, the growth chart was labelled **`-4.1%`** — same agent, same data, same moment. Both are performance claims; only one was behind the gate.

The growth index subtracts each period's *flow* from the equity line so what remains is the agent's doing rather than its owner's. Right idea, and it inherits the flows wholesale: the canary's three phantom opening balances — 10 USDG each, inferred from a balance change, nothing actually deposited — get divided out of a book that never received them.

The chart guarded on `funded`, which is only `contributed !== null`. Three phantom rows satisfy it. It now reads the same durable `contributions_known` verdict the return gate reads, so there is one definition of publishability rather than a second hiding in a component.

The balance history still draws — it is real, and a blank panel teaches nobody anything. What goes is the percentage, plus a sentence saying why. Paper agents are unaffected: they have no real flows by construction, and gating them on real-capital evidence would blank a chart that is honest about a simulation.

## The dry run overstated what a repair unlocks

It reported `PnL publishable true` for the paper accounts whose contributions go to **zero** — where the honest answer is that there is nothing to divide by. Production was never at risk (`rankPnl` refuses on `contributed <= 0`), but a preview read by someone deciding whether to approve a mutation must not overstate what it delivers.

Found by reading the dry run's own fleet output rather than trusting it.

2090 tests, both typechecks clean.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "prepublishOnly": "node cli/prepublish.mjs",
     "dev:web": "cd web && next dev -p 3100 -H 127.0.0.1",
     "dev:worker": "tsx worker/src/index.ts",
-    "test": "tsx --test \"worker/src/**/*.test.ts\" \"web/src/**/*.test.ts\"",
+    "test": "tsx --test \"worker/src/**/*.test.ts\" \"web/src/**/*.test.ts\" \"packages/*/src/**/*.test.ts\"",
     "test:contracts": "cd contracts && npm test",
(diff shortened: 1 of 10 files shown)


millw14:
Superseded by #69 — rebased onto main after #67 squash-merged.

PR #81, opened 2026-09-05
Brain: adaptive depth, its own credential, and core's numbers

millw14:
Follow-on to #76, covering items 3–6.

### Adaptive depth wins — measured on 36 scenarios, 108 runs, $0.2266

| stages | correct | calls | tokens | usd | sec/run |
|---|---|---|---|---|---|
| analysts | 27/36 | 168 | 144,033 | 0.0477 | 4.0 |
| **adaptive** | **28/36** | 210 | 211,561 | 0.0660 | 5.3 |
| full | 27/36 | 338 | 410,376 | 0.1130 | 8.4 |

The full committee scores the **same as analysts-only** at 2.0× the calls and 2.8× the tokens. Adaptive beats both at 58% of the committee's cost.

**Escalation is net positive on the wider set:** 9/36 runs escalated, 2 changed the decision — 1 improved, 0 regressed, 1 neutral. Reasons fired: `opening-a-position` ×6, `large-relative-to-book` ×4. `analysts-disagree` never fired, which is worth watching — the keyword heuristic may be too weak.

Injection resisted 3/3. Refusals cost $0.00 (the gate runs before any model call).

### Its own credential

`LlmConfig` fell back to `GROQ_API_KEY` — the variable the 24 live agents read. Gone. `credential.resolve` takes a key only from `BRAIN_LLM_API_KEY` and **refuses a value identical to any fleet key**: same credential means same quota and same rate limit. Read from a gitignored `.env` (BOM-tolerant, since PowerShell writes one), env var wins, only the last four characters ever logged.

### Core's numbers, not its own

`snapshot.from_canonical` carries `PortfolioSnapshot`: equity is read, never summed; P&L read, never computed; a missing field is a refusal, not a gap to fill. The gate defers to core's `pnl.publishable` outright.

Wired against the canary's **real post-repair snapshot**, which proved its worth immediately — contributions evidenced at 10.000000, and P&L still not publishable because the account sits in epoch 1. The repair fixed the denominator; the epoch gate is a separate rule and still holds.

### One defect the wider suite caught

`gate.assess` defaulted to an epoch floor of 1 while core uses 2, so `epoch_one` returned BUY on a book core calls unmeasurable. A second accounting rule, and the lenient half. `MIN_AUDITABLE_EPOCH` now names it once. Found by measurement, not review — it never fired on the ten hand-written scenarios.

### DIFF
--- a/services/brain/.gitignore
+++ b/services/brain/.gitignore
@@ -3,3 +3,7 @@
 *.jsonl
 __pycache__/
 *.egg-info/
+# The Brain credential lives here and must never be committed. It is a
+# DIFFERENT key from the fleet house key by design — see credential.py.
+.env
+.env.*
--- a/services/brain/.gitignore
+++ b/services/brain/.gitignore
@@ -1,0 +1,5 @@
+.venv/
+.eval/
+*.jsonl
+__pycache__/
+*.egg-info/
--- a/services/brain/brain/budget.py
+++ b/services/brain/brain/budget.py
@@ -1,0 +1,216 @@
+"""
+THE CEILING EXISTS BEFORE THE FIRST CALL.
+
+On 2026-08-31 a background feature nobody had switched on consumed an entire
+200,000-token daily allowance (195,881 used) on a shared house key, and the
+first person to notice was a user whose chat had stopped working. Nothing
+counted. The only defences were an interval timer, a step cap, and flags
+defaulting off — three ways of hoping.
+
+So this module is written before any node that calls a model, and every call
+goes through it. It is not telemetry. A call that would breach the ceiling does
+not happen: `spend()` raises, the run returns a typed refusal, and the partial
+cost is still reported. A truncated decision is not a cheaper decision, it is a
+wrong one, so the run refuses rather than returning what it managed to think.
+
+PER-AGENT CONCURRENCY IS ONE. Not a performance choice — two runs for the same
+agent would double its spend against a shared allowance while producing two
+decisions that disagree, and the trigger layer cannot dedupe what it cannot see.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import json
+import os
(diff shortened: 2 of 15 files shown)


millw14:
**This looks like the same branch proposed twice — recommend closing rather than merging.**

This PR is on `feat/brain-service`, which already merged as #76 ("Brain: how a Merryman thinks, as a service that returns a decision") on 2026-09-05. All 17 files it adds exist on `main` today.

Measured against current `main`, scoped to the brain service, this branch is **+94 / −2,243**. Merging it would not add the service — it would roll back what landed after it:

- `analyst.py` wiring, `AnalystSignal`, `LENS_KEYS` (#85, #88)
- `judge_economics` (#92)
- `fixtures/suite.py`, `tests/test_material.py` (#89, #94)

So in its current state this is not merely redundant, it is destructive.

**What to do:** close it. If anything in "adaptive depth, its own credential, and core's numbers" genuinely never landed, cut a fresh branch from current `main` containing only that delta.

millw14 (commented):

**Verdict: Changes needed before merge; reconcile this older service proposal with current main.**

**[P2] Failed inference attempts do not consume the call budget** — `services/brain/brain/llm.py:127–153`. The comment promises that every attempt is counted, but `budget.record()` runs only after a successful HTTP response. A 429 or HTTP failure retries without incrementing `model_calls`; the next `check_before()` sees the old count. Timeouts can represent provider work that already ran. This permits more outbound attempts than `max_calls` and understates usage during failures. Reserve an attempt before each request, account returned token usage separately, and test a small call budget with a mock provider returning failures.

Validation: source review of the service boundary, credential isolation, budgets and assembly. Python service/provider tests were not run locally. GitHub's app/contracts checks do not substitute for those tests. GitHub reported merge conflicts at review start.

PR #101, opened 2026-09-06
Swap settings + quote-only swap screen (no spend path)

aor-rex:
Settings/UI half of the auto-convert work, carved out per review on #38: **nothing here submits, signs, or moves funds.** Execution (guarded quote path, narrowed native value, pause/policy gating, SQL latch, write-ahead claims) follows in #38.

## Architecture note (important for reviewers)

The `(app)` layout renders `<App/>` (terminal) and drops route children, so `(app)/*` pages are stubs by design. The swap UI therefore lives where all terminal screens live:

- **New `terminal/screens/Swap.tsx`** (quote-only) + `{kind:"swap"}` in `Screen`, wired through `screenForPath`/`pathForScreen` — `/swap` URL-synced like settings/grant/limits.
- **Stub route** `(app)/swap/page.tsx` (metadata + null), same pattern as the settings/grant stubs.
- **Swap entry** in the live You screen fund row (`account-fund` button + `onSwap` prop), next to Add funds / Withdraw.

## What this PR contains

- **Settings fields** (`autoConvertEnabled`/`ReservePct`, `manualSwapWei`/`Id`) with API validation, defaults, and worker resolution. Inert until the execution follow-up wires a consumer.
- **Terminal Settings**: auto-convert toggle + gas reserve + prominent Swap entry in Trading preferences.
- **Swap screen**: live TWAP-guarded quote preview (est. out, minOut, fee, depth/divergence refusals), balances + reserve figures, staged review with the confirm step **disabled** ("execution lands next").
- **Quote API is read-only**: TWAP via `readPoolPrice` + `poolPriceUsable` (no spot fallback — fresh pools wait for history). RPC failures surface as unknown, never zero. Reserve math shared with the future tick via `convertReserve`.
- **Grant native-swap marker + predicate** (read-only surface; no grant mints it yet, no wall change).
- **Census pins** updated for the three added controls (11→12, 12→13, 17→18).

## Verification

- 2,645 tests green, `tsc` clean (worker + web).
- Verified live in the terminal: `/swap` renders the Swap screen with chain balances, re-sign gating, and reserve math; no UserOp, no `execute`, no wall/policy/tick changes.

Supersedes nothing yet: #38 stays open for the execution half and will rebase onto this once it lands.


millw14:
**Needs a rebase, and coordination with #38 and #102.**

CONFLICTING against current `main`. This is one of three open PRs touching the same swap/convert surface: this one (settings UI), #102 (execution) and #38 (the broader original, which overlaps both).

**What to do:** agree the split with #38 and #102 first, so the three do not conflict with each other after rebasing. Then rebase this one and let CI re-run. Since this is described as quote-only with no spend path, please state explicitly in the description which code path could initiate a transfer, so that claim can be checked rather than taken on trust.

---

aor-rex:
Rebased onto current main and split agreed as requested — with #38 closed as superseded, the line is now:

- **#101 (this PR):** settings UI, quote display, terminal screens. Pure interface + validation.
- **#102:** execution — quoted convert leg, TWAP guard, submits, SQL latch, wall rules (including the WETH-pinned native-value split, which lives there, not here).

**Quote-only claim, stated for checking:** this PR's swap screen issues exactly two network calls, both GET — `/api/swap/quote?wei=` and `/api/feed`. Every button mutates local review state (`setReviewing`). The single spend-adjacent surface is the `manualSwapWei`/`manualSwapId` settings handoff (digits-only wei + tight id shape, rejected visibly on malformed input) — inert without a worker consumer, and that consumer lives in #102. No POST, no signer, no submit path in this PR's files.

Rebase resolutions were all additive (both sides kept); test census rebased to measured counts (17 checkboxes / 14 numbers / 23 guards). CI green on this push.

millw14 (commented):

**[P2] An older quote can overwrite the preview for a newer amount** — `web/src/terminal/screens/Swap.tsx:84–88`. The debounce cancels only a timer, not an already running fetch, and every response unconditionally sets quote/loading state. Request A for one amount can resolve after request B for a newer amount, leaving the current input paired with A's amount/output/reserve preview. Clearing the input also does not invalidate A. Use an AbortController or request generation and accept a response only for the current amount/account; clear quote validity immediately when input changes. Test responses arriving in reverse order and after clearing the field.

Scope: this PR keeps execution disabled, so this is a misleading-preview bug rather than a demonstrated spend. The same component is included in #102. Source review and existing green app/contracts/gateway checks; no browser race reproduction was run. GitHub reported merge conflicts at review start.

aor-rex:
Confirmed and fixed — thank you. The race was real: the debounce cancelled only the timer, so an older amount's fetch could resolve after a newer one's and pair the current input with a stale preview.\n\n**Fix:** generation guard (`quote-guard.ts`) — each refresh takes a generation, only the latest writes state; clearing the input retires in-flight fetches and clears the quote immediately. Unit tests cover reverse-order resolution and post-clear arrival.\n\n**Honest boundary, stated plainly:** a full reverse-order *browser* reproduction wasn't run — this screen imports CSS and `next/link`, neither of which the tsx test runner can load, so no jsdom mount exists. The guard rule itself is tested; the wiring is a thin, review-visible application of it. Same evidence standard you hold.\n\n**Also in this push:** honesty census updated for main's trencher checkbox/guard plus this branch's own autoConvert controls (18/15/24, each with its reason).\n\n**Evidence:** full suite green locally; `app`+`contracts`+`gateway` green on the pushed tip. Note this is preview-only scope (no spend path in #101), though the component is shared with #102.

millw14 (commented):

**Needs changes.** The generation helper is useful, but its component wiring still permits stale previews.

**[P2] Invalidate at input change and check after JSON decoding** — `web/src/terminal/screens/Swap.tsx:94–95` and the amount effect below. For a valid amount A changed to another valid amount B, the effect neither invalidates A nor clears its quote: B receives a new generation only when its 400ms timer fires. A may resolve during that interval and remain displayed under B. Separately, generation is checked before `await r.json()`, so a response that passes the check can finish decoding after a newer response and overwrite it. Clear/invalidate on every input change, then check the generation after all awaited decoding immediately before setting state. Add component-level cases for an old response during the debounce and a delayed JSON body; the helper-only tests omit both.

Verification: source and test inspection only; no contributor code or browser session executed. Existing CI and Vercel passed, but this head now conflicts with main. Rebase and rerun required checks. No automatic merge authorization.

PR #102, opened 2026-09-06
Convert execution: guarded TWAP quote, narrowed native value, gated submits, SQL latch

aor-rex:
Execution half of the auto-convert work; settings/UI half is #101 (stacked on it — rebase to `main` once that lands). Nothing here is new surface — it wires the preferences #101 stores to a guarded spend path. Answers the #38 review blockers; supersedes #38's execution content.

**Closes the #38 review blockers:**
- Quote is TWAP-only via `readPoolPrice` + `poolPriceUsable` (spot-fallback path deleted, not fixed).
- Native value split into its own WETH-`EQUAL` router rule (0.5 ETH sealed ceiling); general rule back to 0n; limit threaded through `buildWallPolicies` (0n omits the rule); marker minted in lockstep at signing.
- Both convert paths run below the pause check and pass a scoped policy gate (full `checkPolicy` minus per-trade/daily caps — funding-into-self is not market exposure; see `convertPolicyLimits` rationale).
- Gas-price failure refuses with hourly backoff, worker and API.
- `convert_state` SQL latch (mirrored snapshot + spawn-time seed ferry); file latch removed.
- Failure backoff without marker moves; write-ahead manual-id claims.

**Verification:** 2,658 tests green (new: latch rules/codec, reserve math, scoped-policy proof that unmodified policy misfires, wall split pins, 0n-hatch test), `tsc` clean (worker + web), interactive pass of quote/gating/entry states in the terminal preview. Live-fire on a small amount before merge — not claimed here.

**Deliberate deviations, stated:** scoped (not full) policy gating with rationale in code; per-tick state log quieted to the deposit window. Happy to argue either.

### DIFF
--- a/packages/core/src/wall.ts
+++ b/packages/core/src/wall.ts
@@ -280,7 +280,35 @@ export interface WallOptions {
    * covers.
    */
   ponsAdapterAddress?: Address;
+  /**
+   * Native value the SwapRouter02 NATIVE-INPUT rule may carry, in wei. Default
+   * NATIVE_SWAP_VALUE_LIMIT_WEI (0.5 ETH) so every newly signed wall can run
+   * the ETH→USDG convert; an explicit 0n omits the native rule entirely,
+   * restoring the old no-native wall exactly.
+   *
+   * WHY A SECOND RULE, not a valueLimit on the general one. The general
+   * exactInputSingle rule pins both token legs ONE_OF the asset set with
+   * valueLimit 0n. Hanging native value on it would let the key attach up to
+   * the ceiling to ANY exactInputSingle it may otherwise make — including ones
+   * where the ETH is not the input at all and simply stays with the router.
+   * The native rule instead pins tokenIn EQUAL WETH: it matches exactly the
+   * call shape that needs msg.value (native ETH in, WETH tokenIn, the router
+   * wraps internally), and nothing else. The ceiling bounds a compromised key
+   * to forced bad-price swaps into the account's own holdings — the recipient
+   * is pinned self and there is no transfer permission, so value cannot leave,
+   * only be swapped badly.
+   */
+  nativeSwapValueLimitWei?: bigint;
 }
+
+/**
+ * The ceiling on native value the native-input exactInputSingle rule may
+ * carry — the bound that makes the ETH→USDG convert expressible at all.
+ * 0.5 ETH is far above any conversion a sane reserve split produces and far
+ * below "the account's whole balance" for anything but a dust account, where
+ * the worker's own reserve arithmetic never proposes the op anyway.
+ */
+export const NATIVE_SWAP_VALUE_LIMIT_WEI = 500_000_000_000_000_000n; // 0.5 ETH
 
 /**
  * Owner-added tokens that are safe to seal into a policy.
@@ -359,6 +387,12 @@ export function buildCallPermissions(
     ...extras.map((t) => t.address as Address),
   ];
   const self = { condition: ParamCondition.EQUAL, value: smartAccount } as const;
+  // The native-input ceiling, defaulting to the sealed constant. An explicit
+  // 0n omits the native rule below — the "0n restores the old wall" escape
+  // hatch, and it must stay STRUCTURAL (rule absent) rather than a 0n limit
+  // on a present rule, so old and new walls differ visibly in the permission
+  // list a test can read.
+  const nativeSwapValueLimit = opts.nativeSwapValueLimitWei ?? NATIVE_SWAP_VALUE_LIMIT_WEI;
   // Deduped and lowercased so a list with the same address twice doesn't bloat
   // the on-chain policy, and a case difference can't read as a second address.
   const withdrawals = [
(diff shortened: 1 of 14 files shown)


millw14:
**Mergeable on paper, but please do not merge it as-is — rebase first.**

GitHub reports MERGEABLE, and both required checks passed. That is not sufficient here, because "no textual conflict" is not the same as "safe", and this PR touches the most sensitive files in the repo:

- `packages/core/src/wall.ts` and `worker/src/wall.test.ts`
- `worker/src/orchestrator.ts`, `worker/src/policy.ts`
- `web/src/lib/session.ts`

All of these moved on `main` after this PR was opened — the permission wall changed twice in one day (`77383d8`, `f5e8fe9`), and `orchestrator.ts` changed again in `460c32f`. Its CI result is from 2026-09-06 and was computed against a `main` that no longer exists.

**What to do:** rebase onto current `main`, let CI re-run, and re-read the `wall.ts` diff by hand against the new wall code before merging. A silent semantic conflict on the wall is the one failure mode that would not show up as a conflict marker or a red check.

---

aor-rex:
Rebased onto current main (`46c852d`) and retargeted the PR base to `main` — the 100-commit display was the old stacked base (`8db05a1`); the diff is now the branch's own 14 commits. No code changes beyond the rebase resolutions below.

**Wall hand-read** (against the new wall code, as asked): the general `exactInputSingle` rule still carries `valueLimit 0n` — only the new WETH-pinned split rule carries value (`tokenIn EQUAL WETH`, the exact call shape needing msg.value), and `buildWallPolicies` forwards `nativeSwapValueLimitWei` with no second default (explicit `0n` omits the rule structurally). Main's pons-vault forwarding preserved alongside. No silent semantic conflict found; details in the diff.

**Rebase resolutions** (all additive, both sides kept): ponsClassVaultFactory/assetMode + auto-convert settings fields, tsconfig alias additions, settings-route PUT branches, honesty census rebased to measured counts (17 checkboxes / 14 numbers / 23 guards), App.tsx screen-routing dedup (kept main's `./nav` extraction), ledger-mirror convert-state block, import merges.

**Size-model recalibration** (CI-caught, real): the native rule widens the default wall, so recorded stubs moved (0-token 10,932→12,180; +1,248 fixed, +32/token; 40-token 33,940), permissions 18→19, per-token 512→544, narrow allowance now ~13.08M (anomaly threshold 13M→13.5M), straddle point 6→4. All values measured, none guessed; properties (monotonicity, caps, headroom refusal) unchanged.

**CI state:** `contracts` green. `app` green except one failure I am deliberately not fixing here: `EVERY error declared in PonsSelfTrade.sol is classified` — upstream commit `0a11614` added a contract error without a classifier entry. Proven pre-existing (fails identically with my changes stashed; my diff touches neither `revert.ts` nor `contracts/`). Classifying a new revert reason means deciding retryable-vs-permanent, which is fund-safety judgment for the PR that introduced the error — flagging rather than guessing.

aor-rex:
Rebased onto current `main` per your note — and the rebase itself resolved the red `app`. Diagnosis, with proof: `worker/src/revert.test.ts` fails identically at this branch's old base (`46c852d`, 31/32) and passes on current `main` (38/38, no PR code involved). The gap closed while the branch sat still: `8ffd11c` (wider-error selectors) and `236d0fc` (spend-cap classification) landed after this branch forked, and the test file grew 32→38 around them. So it was a stale base, not this diff — no classifier entries added or needed from my side. Rebase had one conflict (`session.ts` import block: kept both `probeClassFactory` and `GRANT_NATIVE_SWAP`). Local: typecheck clean, full suite green; CI on the rebased tip: `app` and `contracts` both pass.

millw14 (commented):

**[P1] Manual conversion is not durably claimed before money moves** — `worker/src/index.ts:3960–3971` and `persistLatch` at 4200. `submitConvertLeg()` is awaited before `settleManualSwap()` records the request ID. A crash after broadcast/receipt but before settlement leaves the settings request replayable after restart. Moreover, `persistLatch()` starts an asynchronous write without awaiting it and swallows failures, so even moving the current call earlier would not create a durable claim. Atomically claim the ID in durable authoritative storage before any spend; await and fail closed on that write, persist the operation identity, and reconcile ambiguous outcomes instead of replaying them. Cover crashes at broadcast/receipt and storage failures, including hosted redeploy before mirroring.

Validation: 11 focused conversion/latch/reserve tests passed. They exercise in-memory rules, not the crash boundary. The quote UI also inherits the stale-response issue identified in #101. GitHub reported merge conflicts at review start; no live conversion was submitted.

aor-rex:
Confirmed and fixed — thank you. The crash window was real: `submitConvertLeg` ran before any durable claim, and settle's persist was unawaited with failures swallowed, so a crash between broadcast and the DB write left the request replayable (double spend). Notably the module docstring already promised write-ahead claiming — the code just never implemented it.\n\n**Fix:** the id is now recorded + durably persisted (awaited, fail-closed) BEFORE broadcast — persist failure unclaims in memory, keeps the handoff, spends nothing, owner retries with the same id. Settle awaits persist with bounded retries + a loud warning on exhaustion. New `unclaimSwapId` unit test; existing latch/battery suites.\n\n**Found during the rebase this required** (reporting plainly): the rebase had committed conflict markers into `wall.ts` (my `add -A` staged them — my error, fixed in a separate commit, both sides kept), and the branch's second native-input router rule collided with main's new duplicate-permission invariant — two `exactInputSingle` rules on one target make the wall uninstallable (AA23, the f96ddd9 precedent). Merged the ceiling onto the single rule; wall-shape/size/headroom/census tests updated to match (including reverting our +1248 recalibration, which had measured the two-rule wall).\n\n**Evidence:** full suite green locally; `app`+`contracts`+`gateway` green on the pushed tip.

millw14 (commented):

**Needs changes.** Awaiting a claim before submission fixes the immediate local write-after-spend ordering, but the hosted durability boundary from the original review is still open.

**[P1] Persist the pre-spend claim in authoritative hosted storage** — `worker/src/index.ts:4062–4063` and `persistLatchNow` at 4332. Hosted children have DATABASE_URL stripped, so `putConvertState` acknowledges the child-local ledger. The shared database receives that row later through the ledger mirror, while the orchestrator seeds a fresh child from the shared row. A redeploy with a fresh home after broadcast but before mirroring can therefore lose the acknowledged claim and replay the still-pending settings request. Await an atomic claim through an authenticated orchestrator/shared-storage boundary before spending, with an operation identifier and reconciliation of ambiguous outcomes. A local write plus eventual mirroring is not that barrier. Add a test for broadcast → redeploy before mirror with the same request ID, plus storage failure.

Verification: inspected the submit path, store writer, mirror and orchestrator seeding; no live orders or contributor code executed. Existing CI passed on this revision but does not prove that failure boundary, and the branch conflicts with current main. No automatic merge authorization.

PR #107, opened 2026-09-13
feat: stage custom tokens + discovery toggle from Telegram

aor-rex:
Hosted tenants link their own bot via a settings-stored token. This gives them token setup + discovery control from chat, with the dashboard remaining the sole place anything gets permissioned.

**New commands** (all gated by the existing linked-chat allowlist + `CONTROL_KINDS`, like strategy/cap):
- `/addtoken 0x… [SYMBOL]` — validates, reads decimals on-chain, echoes details back, parks for `/confirm` (shared kind-tagged slot, 90s TTL, re-vetted at fire time). A bare `/addtoken yes` also confirms.
- `/discover on|off` — flips the new-pair feed (single-shot, reversible, echoes the Bitquery/Merry Circle credential requirement).
- `/tokens` — lists staged tokens (read-only, ungated).

**Durability, both backings:** hosted writes go to the tenant's sealed store (`get/put`, survives the supervisor's 15s rewrite); self-host patches `settings.json` (established bot-write path, re-read every tick). Tenant = armed grant's smart account, threaded through `TelegramServiceDeps`. Staged tokens surface in Settings → Custom tokens & discovery (same record both sides).

**Boundaries (explicit non-goals):** no grant/wall/key code touched — a staged token confers zero permission until a dashboard re-sign covers it (asserted in tests). No trencher arming, no removals. Model output cannot reach the new commands (`coerceLlmCommand` degrades them to chat — prompt-injection guard, tested).

**Verification:** `tsc -p worker` clean; 173/173 telegram tests pass (11 new); live RPC reads proven (USDG→6, WETH→18, EOA→honest error); 15/15 sandbox wired-flow checks (both backings, re-vetting, gating). Unproven till deployment: literal Telegram round-trip + supervisor sync pickup.


millw14:
**Green on both required checks — held for a product decision, not a technical one.**

`app` and `contracts` both pass, and there is no conflict. Two things to settle before it merges:

1. **Timing.** It edits `worker/src/index.ts`, which is where the class-exit path lives and where active work is happening right now. Please rebase onto current `main` and let CI re-run immediately before merging, rather than relying on the current result.
2. **Direction.** Staging custom tokens by hand pulls against where the product is heading — the class vault exists so an agent can discover and trade new tokens without anyone adding them manually. A per-tenant manual token list may be the right escape hatch anyway, but that is an owner call.

**What to do:** owner decides on (2). If it is a yes, rebase and merge promptly so it does not go stale against `index.ts` again.

---

millw14 (commented):

1. **[P1] Hosted token changes target the wrong settings identity** — `worker/src/index.ts:9193` and `worker/src/telegram/tokens.ts`. `getTenantId()` returns the grant smart account, but the hosted settings API and orchestrator read the authenticated tenant's settings. Those identities differ. The writer can report success while the dashboard/child continues reading a different row. Hosted children also have shared database access stripped, so use an authenticated orchestrator write path keyed to the tenant rather than a direct child-side shared-store write. Verify with distinct tenant and smart-account addresses, then read the value back through the normal hosted settings flow.
2. **[P1] Token confirmation can execute an unrelated pending action** — `worker/src/telegram/interpreter.ts:251`. `/addtoken yes` becomes the generic `confirm` command. If a transfer, kill, or shell action occupies the shared pending slot, this token-specific answer executes that action instead. Preserve the intended action kind and require the pending item to be `addtoken` before confirming it. Add a regression test with a pending transfer and `/addtoken yes`.

Validation: 66 focused token/interpreter tests passed. They mock the tenant identity and explicitly accept the generic confirmation mapping, so neither result proves these integrations safe.

PR #108, opened 2026-09-13
fix: stop re-proposing scout-blocked swaps, quiet repeat refusal pings

aor-rex:
The bot sent the same wall-refusal ping 3× every ~4 min for hours: a strategy proposed the same scout-blocked swap each tick with the budget at 0, each refusal wrote a ledger row, and each row pinged Telegram. The refusal itself was correct policy — the loop around it was the bug.

**Root cause fix (`index.ts` + `quarantine.ts`):** new `scoutGateShut()` mirrors `scoutAllows`' unconditional branches (off / zero budget). The strategy loop and class-entries loop skip deterministically-refused scout buys before decision and ledger — a skip was never attempted, so there is no tape row; a once-per-change feed line is the only record. `scoutBuyUnpriceable()` extracted and shared between the gate and the wall's context builder so the two answers can't drift (class-token history preserved with the code).

**Notifier dedup (`notifier.ts`, defense in depth):** 1h per-rule cooldown on refusal pings, mirroring the existing condition-alert cooldown. First occurrence always fires, a different rule always fires, landed/paper rows untouched. Cursor still advances on skipped rows. Log lines with row ids on every send.

**Untouched:** `policy.ts` strictness, grants/walls/keys, digest behavior, all other notification types.

**Verification:** `tsc -p worker` clean; full worker suite 3148/3148 green (incl. new gate + dedup tests and the updated custody source-structure test). Live proof needs one tick cycle post-deploy.


millw14:
**Needs a rebase — CONFLICTING against current `main`.**

One commit, 6 files. The scout and refusal paths have both changed on `main` since this was opened, which is where the conflict is coming from.

**What to do:** rebase and let CI re-run. Since the fix is about not re-proposing a blocked swap and not repeating refusal pings, please make sure there is a test that fails without the change — a de-duplication fix is easy to land in a form that quietly stops surfacing real refusals too.

---

aor-rex:
Rebased onto current main and reworked for #110, which landed while this was open and refactored exactly the code this touches.

**The important change since the last push:** my pre-proposal gate no longer carries its own target-only class rule — `scoutBuyUnpriceable` now delegates its class leg to `scoutFlagsFor` (same function the wall uses), so gate and wall cannot drift. The old rule would have misjudged class sells exactly the way #110 fixed; the delegation keeps that fix intact in the gate path, including on sells, which must never be gate-skipped. Custody test updated to pin the delegation.

**Dropped in the rework:** my notifier-side per-rule cooldown is gone. Main now pings refusals every tick deliberately (remedy shown once per rule change), so suppression would have fought that design. The gate alone stops the incident (no rows written → nothing to ping); kept only the `[notify]` log lines.

**Verification:** worker + web typechecks clean, targeted suites green (quarantine, custody, class-side, notifier, policy, curve-budget), full worker suite 3549/3550 — the single failure is the pre-existing upstream Pons classifier gap (`0a11614` added a contract error with no classifier entry; fails on clean main, untouched by this diff).

aor-rex:
Rebased onto current main and reworked for #110, which landed while this was open and refactored exactly the code this touches.

**The important change since the last push:** the gate helper no longer carries its own target-only class rule — `scoutBuyUnpriceable` now delegates its class leg to `scoutFlagsFor` (same function the wall uses), so gate and wall cannot drift. The old rule would have misjudged class sells exactly the way #110 fixed; the delegation keeps that fix intact in the gate path, including on sells, which must never be gate-skipped. Custody test updated to pin the delegation, plus a new composition test proving a sell is never gate-shut.

**Dropped in the rework:** the notifier-side per-rule cooldown is gone. Main now pings refusals every tick deliberately (remedy shown once per rule change), so suppression would have fought that design. The gate alone stops the incident (no rows written → nothing to ping); kept only the `[notify]` log lines.

**Verification:** worker + web typechecks clean, targeted suites green (quarantine, custody, class-side, notifier, policy, curve-budget), full worker suite 3694/3695 — the single failure is the pre-existing upstream Pons classifier gap (`0a11614` added a contract error with no classifier entry; fails on clean main, untouched by this diff).

millw14 (commented):

**Verdict: Changes needed before merge; integration conflicts also need resolution.**

**[P2] The refusal deduplicator repeats forever for multiple tokens** — `worker/src/index.ts:10421` and the equivalent class-entry loop at 10525. Only one `lastScoutSkipKey` is retained. With refused proposals A then B each tick, the stored key ends at B; next tick A differs, then B differs, so both warnings are emitted every tick. This reproduces the noise the PR is intended to remove. Track keys per token/configuration or summarize the blocked set, and test repeated ticks containing at least two refused assets plus a configuration change.

Validation: 44 focused quarantine/custody tests passed. The helper rules preserve exits; the repeated multi-intent loop is not covered. GitHub reported merge conflicts at review start.

PR #134, opened 2026-09-19
feat: owner-named withdrawal doors sealed at grant signing

aor-rex:
## Problem

The withdrawal allowlist has been enforcement without registration: the wall (`ONE_OF` pin) and the policy mirror knew how to judge doors, but no signer ever sealed any. Consequences, all live today:

- grants signed now carry NO transfer permission — chat sends are dead on arrival for every new user;
- pre-allowlist grants still send ANYWHERE (grandfathered free-form) — the exact shape a scammer needs;
- the safe middle (send only to pre-approved doors) has no on-switch.

## Change

- `packages/core/src/grant.ts` — `grantWithdrawals?: {name, address}[]`, mirroring the `grantTokens` precedent. Absent = pre-allowlist (grandfathered); `[]` = signed with no doors. `grantHasTransfer` honours sealed doors alongside the grandfathered grants; the marker alone is still never evidence.
- `web/src/lib/session.ts` — `MintOptions.withdrawalAddresses` (max 5, names required) threaded through all four mint paths (appended-last per convention) into `wallOpts` (so the signable gas-ceiling check accounts for door-driven wall growth), the `GRANT_TRANSFER` marker (lockstep), and the always-recorded `grantWithdrawals`. `normalizeWithdrawals` (pure, tested) + throwing wrapper: charset/reserved-word/`0x`-lookalike guards, lowercase/dedupe, blank-row skipping.
- `worker/src/limits.ts` — three-state mirror: sealed doors / `[]` / absent. (Caught during implementation: the naive marker check would have made door-grants lenient — fixed to branch on the recorded list first.)
- `worker/src/index.ts` + `telegram/service.ts` + `telegram/interpreter.ts` — `substituteWithdrawalNames` resolves door labels to addresses at the single choke point before slash parse AND the LLM classifier (voice included), so the verbatim-address rules still see a real address. Slash heads and hex containment guarded; reserved names rejected at registration.
- `web/src/terminal/screens/CreateAgent.tsx` — doors section in the limits step with live validation, add/remove, and a plain-language summary ("Money can leave only to: …" / "No doors: the agent won't be able to send money anywhere").

## Explicit non-goal

No chat-side registration path, ever. The wall cannot grow by itself: doors enter only through an owner signature. Chat resolves names; it never approves them.

## Tests

- New: `normalizeWithdrawals` matrix (web), `substituteWithdrawalNames` matrix (interpreter), sealed-doors mirror cases (transfer-mirror), doors row + expectation fix (wall-battery).
- Existing policy allowlist matrix untouched and green.
- Local: `npm ci` + typecheck (worker/web/browser/sdk) + full suite (5098 pass / 0 fail).

Draft PR — review welcome whenever; no rush.


millw14 (commented):

1. **[P2] Empty withdrawal lists falsely fail the wall battery** — `worker/src/wall-battery.ts:210`. Every newly signed grant records `grantWithdrawals: []` when no destinations were selected. This branch expects `transfer-recipient-allowlist`, but the policy correctly returns `transfer-not-permitted`. I reproduced `held: false` on the first battery case for this ordinary safe grant. Check transfer capability/empty lists before selecting the expected rule and add a test using the actual new-grant shape.
2. **[P2] Name substitution rewrites transfer amounts and unrelated commands** — `worker/src/telegram/interpreter.ts:195`. Normalization accepts a destination named `50`; I reproduced `send 50 to cold wallet` becoming `send 0x1111111111111111111111111111111111111111 to cold wallet`. Substitution applies to the entire message, not the recipient slot. Restrict resolution to transfer destinations and reject ambiguous names; preserve amounts and unrelated text.

Validation: 80 focused tests passed. The two additional reproductions expose cases those tests omit. Browser signing and an on-chain withdrawal were not exercised.

aor-rex:
Both reproductions confirmed against the code and fixed — thank you.\n\n**P2-2 (amount rewriting) — my bug, as described.** Nothing rejected a door named `"50"`, and whole-word substitution can't tell the label from the amount. Fixed in `normalizeWithdrawals": names must now contain at least one letter (reserved-word and charset rules already covered the rest). Tests cover pure numerics, decimals, and the `door50`-style names that stay legal.\n\n**P2-1 (empty-list expectation) — my builder logic, confirmed.** The policy returns `transfer-not-permitted` for `[]` before ever reaching the allowlist check, but my battery expectation mapped every defined list to `transfer-recipient-allowlist` — a safe empty-door grant would have reported BREACH. Fixed the expectation to branch on empty vs non-empty, with the new test using the actual new-grant shape (marker + `[]`) you asked for.\n\n**Evidence:** targeted doors/mirror/battery suites green; full suite green locally; `app`+`contracts`+`gateway` green on the pushed tip.

millw14 (commented):

**Needs changes.** The empty-withdrawal-list battery expectation is corrected, and rejecting bare-number labels fixes the reported `50` example. The unrelated-command part of the substitution finding remains.

**[P2] Resolve door names only in the recipient position** — `worker/src/telegram/interpreter.ts:185–195` and its unconditional service call before `parseSlash`. A valid door named `cold wallet` still rewrites `/type cold wallet` into typing its address; a door named `TSLA` rewrites an unrelated token request. Rejecting numeric names does not fix whole-message replacement. Parse the intended action first and resolve names only in a transfer recipient field. Test that `/type`, shell text and token requests preserve their original arguments.

Verification: read the updated implementation and regression tests; did not execute contributor code or submit a withdrawal. Existing app/contracts/gateway checks passed, but this revision conflicts with current main and Vercel remains unauthorized. Wallet permission changes require Milla's review; no automatic merge authorization.

